PR #226, opened 2023-10-06
NEOS-118: adds in base transformer table to the front end

evisdrenova:


backend/internal/nucleusdb/json-models/models.go
```diff
@@ -132,29 +132,158 @@ type JobMapping struct {
...
+	Value  string
+	Config *TransformerConfigs
+}
+
+type TransformerConfigs struct {
+	EmailConfig *EmailConfigs
+	FirstName   *FirstNameConfig
+	Uuidv4      *Uuidv4Config
+	PhoneNumber *PhoneNumberConfig
+	Passthrough *PassthroughConfig
```
nickzelei, line 174: Since you already did this, it's fine - but we can most likely get away with only needing configs for transformers that actually require it. Having them be empty configs isn't really doing much.

backend/internal/nucleusdb/json-models/models.go
```diff
@@ -132,29 +132,158 @@ type JobMapping struct {
...
+		t.Config = &TransformerConfigs{
+			Uuidv4: &Uuidv4Config{},
+		}
+	case *mgmtv1alpha1.TransformerConfig_PhoneNumberConfig:
+		t.Value = tr.Value
+		t.Config = &TransformerConfigs{
+			PhoneNumber: &PhoneNumberConfig{},
+		}
+	default:
+		return fmt.Errorf("invalid config")
```
nickzelei, line 224: If a transformer doesn't have a config, I think the default should just be: 
```
t.Value = tr.Value
t.Config = &TransformerConfigs{}
```
Then not ever transformer needs a config.

backend/internal/nucleusdb/json-models/models.go
```diff
@@ -132,29 +132,158 @@ type JobMapping struct {
...
+		return &mgmtv1alpha1.Transformer{
+			Value: t.Value,
+			Config: &mgmtv1alpha1.TransformerConfig{
+				Config: &mgmtv1alpha1.TransformerConfig_UuidConfig{
+					UuidConfig: &mgmtv1alpha1.Uuidv4{},
+				},
+			},
+		}
+	default:
+		return nil
```
nickzelei, line 283:
```
return &mgmtv1alpha1.Transformer{ Value: t.Value }
```

backend/protos/mgmt/v1alpha1/job.proto
```diff
@@ -286,9 +278,7 @@ message GetTransformersResponse {
 }
 
 message Transformer {
-  string title = 1;
   string value = 2;
```
nickzelei, line 281:
```suggestion
  string value = 1;
```

backend/protos/mgmt/v1alpha1/job.proto
```diff
@@ -286,9 +278,7 @@ message GetTransformersResponse {
...
-  string description = 3;
   TransformerConfig config = 4;
```
nickzelei, line 282:
```suggestion
  TransformerConfig config = 2;
```

nickzelei (commented):
Left a few minor comments.

Is there any way to make this look more uniform?
image

There also seems to be a lot of lag when I select a transformer, I wonder what is causing such a performance problem?

backend/internal/nucleusdb/json-models/models.go
evisdrenova: yeah i left them blank for right now but intending on filling them up later, likely these will all have some configs besides the passthrough one

PR #446, opened 2023-10-27
improve job runs page

alishakawaguchi:
Fixed deleting job run
added delete + cancel buttons to individual job run page
fixed polling on individual job run page
reordered recent job runs on job overview page to show most recent runs at top of table
added polling to job overview page when a job run is triggered
fixed transformers object missing from get job response


nickzelei:
NEOS-250

PR #500, opened 2023-11-06
Docusaurus3.0

nickzelei:


nickzelei:
Can't merge until: https://github.com/facebook/docusaurus/issues/9483 is closed.

PR #532, opened 2023-11-08
invites table migration

alishakawaguchi:


backend/hack/migration-add.sh
```diff
@@ -14,4 +14,4 @@ SCRIPT_DIR=$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)
 
 # use epoch time so unlikely to clash
 set -v
-migrate create -ext sql -dir "${SCRIPT_DIR}/../migrations" "$1"
+migrate create -ext sql -dir "${SCRIPT_DIR}/../sql/postgresql/schema" "$1"
```
nickzelei, line 17: whoop sorry, missed this ref update.

backend/sql/postgresql/schema/20231108001411_invites.up.sql
```diff
@@ -0,0 +1,19 @@
...
+  email varchar NOT NULL,
+  token varchar NOT NULL DEFAULT gen_random_uuid(),
+  accepted boolean DEFAULT false,
+	created_at timestamp NOT NULL DEFAULT now(),
+	updated_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  expires_at timestamp NOT NULL,
+	CONSTRAINT invites_pkey PRIMARY KEY (id),
+  CONSTRAINT fk_invites_accounts_id FOREIGN KEY (account_id) REFERENCES neosync_api.accounts(id) ON DELETE CASCADE,
+  CONSTRAINT fk_invites_users_id FOREIGN KEY (id) REFERENCES neosync_api.users(id) ON DELETE CASCADE,
+  CONSTRAINT invites_email_account_id UNIQUE (email, account_id)
```
nickzelei, line 14: I don't know if we want this constraint. If an email gives invited but the invitation expires, they should be able to receive another invite without having to delete the previous one.

backend/sql/postgresql/schema/20231108001411_invites.up.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE IF NOT EXISTS neosync_api.invites (
+	id uuid NOT NULL DEFAULT gen_random_uuid(),
+  account_id uuid NOT NULL,
+  sender_user_id uuid NOT NULL,
```
nickzelei, line 4: I'm thinking this should be NULL and the FK constraint should set to null if the user_id is deleted. That way a user could be removed, but the invite is retained.
Any thoughts?

alishakawaguchi: yeah that makes sense

nickzelei (approved):
🚢

PR #548, opened 2023-11-09
Use devcontainers CLI directly

jacobwgillespie:
The `devcontainers/ci` action assumes that multi-platform image builds aren't possible and injects an `--output` flag to download the image as a tarball instead. This isn't the case with Depot, as it can build multi-platform images directly.

This PR adapts the workflow to call the `devcontainer` CLI directly, so that the build and push can function as expected.


nickzelei (approved):
Amazing! Thank you so much for helping with this.

PR #764, opened 2023-12-06
Updates backend,worker to work with docker compose watch

nickzelei:


compose.yml
```diff
@@ -86,7 +86,7 @@ services:
     develop:
       watch:
         - path: backend/bin
-          action: rebuild
```
nickzelei, line 89: The Docker docs say that `rebuild` is the same as `docker compose up --build` but the latter command works and this rebuild does not.
It results in saying the application could not be restarted

However, `sync+restart` seems to be doing the trick since we sync the binary and restart the process.

backend/dev/build/Dockerfile.dev
```diff
@@ -5,10 +5,11 @@ RUN mkdir /app
 
 WORKDIR /app
 
-COPY /bin/mgmt /app
 COPY /sql/postgresql/schema/ /app/migrations/
 
 RUN echo 'NUCLEUS_ENV=dev' > /app/.env
 COPY .env.dev /app
 
+COPY /bin/mgmt /app
```
nickzelei, line 13: Want this to be last so that everything else is cached.

PR #956, opened 2023-12-22
Nick/neos 546 move over to contentlayer for docs and marketing sites3

nickzelei:


nickzelei:
Cancelling this as contentlayer is unmaintained due to Netlify merger with company that maintained this project.

We use contentlayer for our blogs and that is working fine, but we shouldn't invest any more into this technology.

PR #988, opened 2023-12-28
NEOS-570: custom code transformer - updated protos

evisdrenova:


backend/protos/mgmt/v1alpha1/transformer.proto
```diff
@@ -252,6 +253,10 @@ message Passthrough {}
 
 message Null {}
 
+message Javascript {
```
nickzelei, line 256: Need to differentiate between Generate or Transform

backend/services/mgmt/v1alpha1/transformers-service/transformers.go
```diff
@@ -516,6 +517,17 @@ func (s *Service) GetSystemTransformers(
 					},
 				},
 			},
+			{
+				Name:        "Javascript",
+				Description: "Write custom javascript to transform data",
+				DataType:    "string",
```
nickzelei, line 523: Not really sure this data type makes sense here. The javascript transformer/generator should work with any data type.

backend/services/mgmt/v1alpha1/transformers-service/transformers_test.go
```diff
@@ -39,7 +39,7 @@ func Test_GetSystemTransformers(t *testing.T) {
 
 	assert.NoError(t, err)
 	assert.NotNil(t, resp)
-	assert.Equal(t, 37, len(resp.Msg.GetTransformers())) // the number of system transformers
+	assert.Equal(t, 38, len(resp.Msg.GetTransformers())) // the number of system transformers
```
nickzelei, line 42: Is this a valid test case? It's going to fail any time we add or remove one.

At best this should just assert that `GetTransformers` does not return an empty list if you want to ensure that it always returns some.

worker/internal/benthos/config.go
```diff
@@ -58,7 +58,12 @@ type PipelineConfig struct {
 }
 
 type ProcessorConfig struct {
-	Mutation string `json:"mutation" yaml:"mutation"`
+	Mutation   string           `json:"mutation" yaml:"mutation"`
+	Javascript JavascriptConfig `json:"javascriptConfig" yaml:"javascriptConfig"`
```
nickzelei, line 62: Now that we have two different types of processors, this should be updated to omitempty for both of them.

That way if someone provides only mutations, or only javascript configs, the other does not appear in the yaml.

```suggestion
	Mutation   *string           `json:"mutation,omitempty" yaml:"mutation,omitempty"`
	Javascript *JavascriptConfig `json:"javascriptConfig,omitempty" yaml:"javascriptConfig,omitempty"`
```

Also, I don't think that json is correct. According to benthos the serialized prop should j ust be `javascript`.

image

Which makes this:

```suggestion
	Mutation   *string           `json:"mutation,omitempty" yaml:"mutation,omitempty"`
	Javascript *JavascriptConfig `json:"javascript,omitempty" yaml:"javascript,omitempty"`
```

worker/pkg/workflows/datasync/activities/activities.go
```diff
@@ -168,14 +168,14 @@ func (b *benthosBuilder) buildBenthosSqlSourceConfigReponses(
...
 		if err != nil {
 			return nil, err
 		}
-		if mutation != "" {
-			bc.StreamConfig.Pipeline.Processors = append(bc.StreamConfig.Pipeline.Processors, neosync_benthos.ProcessorConfig{
-				Mutation: mutation,
-			})
+
+		if processorConfig.Mutation != "" || processorConfig.Javascript.Code != "" {
+			bc.StreamConfig.Pipeline.Processors = append(bc.StreamConfig.Pipeline.Processors, *processorConfig)
```
nickzelei, line 178: let's make sure this test case gets covered since the logic here is changing.

worker/pkg/workflows/datasync/activities/activities.go
```diff
@@ -563,32 +563,61 @@ func getMysqlDsn(
...
+
 		}
 	}
-	return strings.Join(pieces, "\n"), nil
+
+	pc := &neosync_benthos.ProcessorConfig{
+		Mutation: strings.Join(mutations, "\n"),
+		Javascript: neosync_benthos.JavascriptConfig{
+			Code: strings.Join(javascript, "\n"),
+		},
```
nickzelei, line 606: This should be conditional here based on the presence of these configs.

So essentially something like this:

```
mutation := strings.Join(mutations, "\n")
js := strings.Join(javascript, "\n")

pc := &neosync_benthos.ProcessorConfig{}
if len(mutation) > 0 {
  pc.Mutation = &mutation
}
if len(js) > 0 {
  pc.Javascript = &js
}
```

worker/pkg/workflows/datasync/activities/activities_test.go
```diff
@@ -310,6 +310,154 @@ func Test_buildProcessorMutation(t *testing.T) {
...
+		"fake-stage-url": pg_queries.NewMockDBTX(t),
+	}
+	pgquerier := pg_queries.NewMockQuerier(t)
+	mysqlcache := map[string]mysql_queries.DBTX{}
+	mysqlquerier := mysql_queries.NewMockQuerier(t)
+
+	bbuilder := newBenthosBuilder(pgcache, pgquerier, mysqlcache, mysqlquerier, mockJobClient, mockConnectionClient, mockTransformerClient)
+	ctx := context.Background()
+
+	mockTransformerClient.On(
```
nickzelei, line 332: This test could be a lot simpler if it instead just defined a standard javascript system transformer instead of a user defined one. One less thing to mock.
We have other tests that should be testing the user defined transformer.

worker/pkg/workflows/datasync/activities/activities_test.go
```diff
@@ -310,6 +310,154 @@ func Test_buildProcessorMutation(t *testing.T) {
...
+		Transformer: &mgmtv1alpha1.UserDefinedTransformer{
+			Id:          "456",
+			Name:        "stage",
+			Description: "description",
+			DataType:    "string",
+			Source:      "javascript",
+			Config: &mgmtv1alpha1.TransformerConfig{
+				Config: &mgmtv1alpha1.TransformerConfig_JavascriptConfig{
+					JavascriptConfig: &mgmtv1alpha1.Javascript{
+						Code: `(()=>{var payload = benthos.v0_msg_as_structured();payload.value+=" helloee";})();`,
```
nickzelei, line 348: I'm pretty confused by this as it's not really at all what we talked about.

The point was to abstract away the benthos stuff from the user, where as this is having them define everything.

Maybe we can talk about this more offline.

worker/pkg/workflows/datasync/activities/activities_test.go
```diff
@@ -310,6 +310,154 @@ func Test_buildProcessorMutation(t *testing.T) {
...
+			Config: &mgmtv1alpha1.TransformerConfig_UserDefinedTransformerConfig{
+				UserDefinedTransformerConfig: &mgmtv1alpha1.UserDefinedTransformerConfig{
+					Id: "456",
+				},
+			},
+		}}},
+	})
+
+	assert.NoError(t, err)
+	assert.Equal(t, res.Javascript.Code, `(()=>{var payload = benthos.v0_msg_as_structured();payload.id+=" helloee";})();`)
```
nickzelei, line 366: How does this work if there are multiple job mappings that define a javascript processor? Would be great to see a test for the expected output here.

nickzelei (commented):
Left a number of comments. Please review. Probably want to chat more about the javascript code implementation.

backend/services/mgmt/v1alpha1/transformers-service/transformers.go
evisdrenova: updating to "any"

backend/services/mgmt/v1alpha1/transformers-service/transformers_test.go
evisdrenova: update

worker/internal/benthos/config.go
evisdrenova: ah yeah good catch, it's just javascript

backend/services/mgmt/v1alpha1/transformers-service/transformers.go
```diff
@@ -54,7 +56,7 @@ const (
 	Passthrough               Transformation = "passthrough"
 	Null                      Transformation = "null"
 	Invalid                   Transformation = "invalid"
-	Javascript                Transformation = "javascript"
+	Javascript                Transformation = "transform_javascript"
```
nickzelei, line 59: Probably want to change this property to `TransformJavascript` as well as I think we will also want a javascript generator.

backend/services/mgmt/v1alpha1/transformers-service/transformers.go
```diff
@@ -723,3 +725,33 @@ func (s *Service) IsTransformerNameAvailable(ctx context.Context, req *connect.R
...
+	}
+
+	return connect.NewResponse(&mgmtv1alpha1.ValidateUserJavascriptCodeResponse{
+		Valid: true,
+	}), nil
+}
+
+func constructJavascriptCode(jsCode, col string) string {
+	if jsCode != "" {
+		return fmt.Sprintf(`(()=>{function fn1(value){%s};const input = benthos.v0_msg_as_structured();const output = { ...input };output["%[2]s"] = fn1(input["%[2]s"]);benthos.v0_msg_set_structured(output);})();`, jsCode, col)
```
nickzelei, line 753: Can we simplify this function and move it to multiple lines so it's easier to read? It's a multiline string anyways.
```
return fmt.Sprintf(`(() => {
    function fn1(value) {
      %s
    }
  })
`, userCode)
```

Does it need to be more than that for a simple test?

backend/services/mgmt/v1alpha1/transformers-service/transformers_test.go
```diff
@@ -420,6 +420,43 @@ func Test_IsConnectionNameAvailable_False(t *testing.T) {
 	assert.Equal(t, false, resp.Msg.IsAvailable)
 }
 
+func Test_ValidateUserJavascriptCode_True(t *testing.T) {
+	m := createServiceMock(t)
+
+	code := `var payload = value+=" hello";return payload;`
```
nickzelei, line 426: let's see an example where the script is actually on multiple lines

worker/pkg/workflows/datasync/activities/activities.go
```diff
@@ -599,16 +601,34 @@ func (b *benthosBuilder) buildProcessorConfig(ctx context.Context, cols []*mgmtv
...
-		Javascript: neosync_benthos.JavascriptConfig{
-			Code: strings.Join(javascript, "\n"),
-		},
+	mutationStr := strings.Join(mutations, "\n")
+
+	pc := &neosync_benthos.ProcessorConfig{}
+	if len(mutationStr) > 0 {
+		pc.Mutation = &mutationStr
+	}
+	if len(javascript) > 0 && !sliceContainsEmptyString(javascript) {
```
nickzelei, line 610: if youre worried about empty strings, then that should be handled above at the construction step on line 587.

worker/pkg/workflows/datasync/activities/activities.go
```diff
@@ -582,10 +584,10 @@ func (b *benthosBuilder) buildProcessorConfig(ctx context.Context, cols []*mgmtv
...
 			}
 
-			if col.Transformer.Source == "javascript" {
-				// parse the js code and replace the key param word which is 'value' with the column name
-				parsed := parseJavascriptForColumnName(col.Transformer.Config.GetJavascriptConfig().GetCode(), "value", col.Column)
-				javascript = append(javascript, parsed)
+			if col.Transformer.Source == "transform_javascript" {
+				// construct the js code
+				js := constructJavascriptCode(col.Transformer.Config.GetJavascriptConfig().GetCode(), col.Column)
+				javascript = append(javascript, js)
```
nickzelei, line 590: don't append if the js string is empty

worker/pkg/workflows/datasync/activities/activities.go
```diff
@@ -620,6 +640,14 @@ func parseJavascriptForColumnName(jsCode, targetWord, replacementWord string) st
 	return strings.ReplaceAll(jsCode, targetWord, replacementWord)
 }
 
+func constructJavascriptCode(jsCode, col string) string {
+	if jsCode != "" {
+		return fmt.Sprintf(`function fn1(value){%s};const input = benthos.v0_msg_as_structured();const output = { ...input };output["%[2]s"] = fn1(input["%[2]s"]);benthos.v0_msg_set_structured(output)`, jsCode, col)
```
nickzelei, line 645: Please use newlines here. this is not very readable.

nickzelei: Also this still isn't quite right. You need to take all of the columns that are using a js processor and build a single script that processes all of them at once. We dont want to split that across N processors. Minimize benthos overhead by having a single processor for handling javascript.

worker/pkg/workflows/datasync/activities/activities_test.go
```diff
@@ -206,6 +210,114 @@ output:
...
+        code: |
+          (() => { 
+          function fn1(value){
+          var a = value + "test";
+          return a };
+          const input = benthos.v0_msg_as_structured();
+          const output = { ...input };
+          output["name"] = fn1(input["name"]);
+          benthos.v0_msg_set_structured(output);
+          })();
```
nickzelei, line 247: We need a test here that shows what this looks like with multiple columns being transformed.

Also it would be great if this were formatted better.

worker/pkg/workflows/datasync/activities/activities_test.go
```diff
@@ -206,6 +210,114 @@ output:
...
+func Test_Sync_Run_Success_Javascript(t *testing.T) {
+	testSuite := &testsuite.WorkflowTestSuite{}
+	env := testSuite.NewTestActivityEnvironment()
+
+	activities := &Activities{}
+	env.RegisterActivity(activities)
+
+	// benthos writes to stdout, need to capture that to evaluate the returned value, so we create a pipe to capture what is written to stdout
+	r, w, _ := os.Pipe()
+	os.Stdout = w
```
nickzelei, line 226: idk about this. Looks like you are using the global Stdout to be this os.Pipe for a single test.
How does this work for the rest of the tests?

backend/services/mgmt/v1alpha1/transformers-service/transformers.go
evisdrenova: was trying to be consistent but simplified it

backend/services/mgmt/v1alpha1/transformers-service/transformers_test.go
evisdrenova: updated

backend/services/mgmt/v1alpha1/transformers-service/transformers.go
nickzelei: This doesnt look like it was resolved.

worker/internal/benthos/transformers/generate_ssn_test.go
```diff
@@ -10,11 +11,15 @@ import (
...
+		assert.NoError(t, err)
+		fmt.Println("res", res)
+	}
 
-	assert.NoError(t, err)
-	assert.IsType(t, "", res, "The actual value type should be a string")
-	assert.True(t, isValidSSN(res), `The returned ssn should follow this regex = ^\d{3}-\d{2}-\d{4}$)`)
+	// assert.NoError(t, err)
+	// assert.IsType(t, "", res, "The actual value type should be a string")
+	// assert.True(t, isValidSSN(res), `The returned ssn should follow this regex = ^\d{3}-\d{2}-\d{4}$)`)
```
nickzelei, line 22: what's going on here?

worker/pkg/workflows/datasync/activities/activities.go
```diff
@@ -563,32 +565,79 @@ func getMysqlDsn(
...
+		t.Source != "generate_default"
+}
+
+func parseJavascriptForColumnName(jsCode, targetWord, replacementWord string) string {
+	return strings.ReplaceAll(jsCode, targetWord, replacementWord)
+}
+
+func constructJavascriptCode(jsCode, col string) string {
+	if jsCode != "" {
+		return fmt.Sprintf(`function fn%[2]s(value){%s};const input = benthos.v0_msg_as_structured();const output = { ...input };output["%[2]s"] = fn%[2]s(input["%[2]s"]);benthos.v0_msg_set_structured(output)`, jsCode, col)
```
nickzelei, line 637: Can this be constructed on multiple lines so that it is easier to read?

worker/pkg/workflows/datasync/activities/activities.go
evisdrenova: it's producing this as an output:

`processors:
    - javascript:
        code: |-
            (() => {
            function fnname(value) {
                var payload = value+=" hello";return payload;
            };
            const input = benthos.v0_msg_as_structured();
            const output = { ...input };
            output["name"] = fnname(input["name"]);
            benthos.v0_msg_set_structured(output);


            function fnfirst_name(value) {
                var payload = value+=" firstname";return payload;
            };
            const input = benthos.v0_msg_as_structured();
            const output = { ...input };
            output["first_name"] = fnfirst_name(input["first_name"]);
            benthos.v0_msg_set_structured(output);
            })();`

where both transformers for different columns (name & first_name) are in the same javascript processor.

worker/pkg/workflows/datasync/activities/activities_test.go
evisdrenova: added a test to the benthos_builder_test file called `Test_ProcessorConfigMultiJavascript`

worker/pkg/workflows/datasync/activities/activities_test.go
evisdrenova: i was trying to come up with a way to actually test the transformed value output and since benthos writes to stdout for `codec:lines` i'm piping that and then getting the resulting value. I wasn't able to find anywhere else where we were actually processing the benthos config and checking the output vs. just the formatting and structure of the benthos config. Open to other ideas.

worker/pkg/workflows/datasync/activities/activities.go
```diff
@@ -634,7 +633,15 @@ func parseJavascriptForColumnName(jsCode, targetWord, replacementWord string) st
...
-		return fmt.Sprintf(`function fn%[2]s(value){%s};const input = benthos.v0_msg_as_structured();const output = { ...input };output["%[2]s"] = fn%[2]s(input["%[2]s"]);benthos.v0_msg_set_structured(output)`, jsCode, col)
+		return fmt.Sprintf(`
+function fn%s(value){
+  %s
+};
+const input = benthos.v0_msg_as_structured();
+const output = { ...input };
+output["%s"] = fn%s(input["%s"]);
+benthos.v0_msg_set_structured(output);
+`, col, jsCode, col, col, col)
```
nickzelei, line 644: this might be fine for now but the reason why in my examples I used numbers for the fn name is that idk if all json keys are compatible as function names. maybe we leave it for now until we find an issue.

worker/pkg/workflows/datasync/activities/benthos-builder_test.go
```diff
@@ -1929,126 +1930,93 @@ func Test_ProcessorConfigEmptyJavascript(t *testing.T) {
...
+						benthos.v0_msg_set_structured(output1);
+	
+						function fnfirst_name(value) {
+							var payload = value += " firstname";
+							return payload;
+						};
+						const input2 = benthos.v0_msg_as_structured();
+						const output2 = { ...input2 };
+						output2["first_name"] = fnfirst_name(input2["first_name"]);
+						benthos.v0_msg_set_structured(output2);
```
nickzelei, line 2016: you'll need to be smarter here about the script generation. this is going to be really memory intensive with the number of inputs because we're creating copies of the object every time.

Can this be consolidated down to this:

```suggestion
						function fnname(value) {
							var payload = value += " hello";
							return payload;
						};	
						function fnfirst_name(value) {
							var payload = value += " firstname";
							return payload;
						};
						const input = benthos.v0_msg_as_structured();
						const output = { ...input };
						
		                                 output["name"] = fnname(input["name"]);
						output["first_name"] = fnfirst_name(input["first_name"]);
						benthos.v0_msg_set_structured(output2);
```

worker/pkg/workflows/datasync/activities/benthos-builder_test.go
```diff
@@ -1929,126 +1930,93 @@ func Test_ProcessorConfigEmptyJavascript(t *testing.T) {
...
+	a, _ := yaml.Marshal(res[0].Config.Pipeline)
+	fmt.Println("res", string(a))
+
+	assert.Equal(
+		t,
+		strings.TrimSpace(`
+			- javascript:
+				code: |
+					(() => {
+						function fnname(value) {
```
nickzelei, line 2000: minor format but if the col name is going to be a part of the function,  let's at least separate it by an underscore `fn_%s`

worker/pkg/workflows/datasync/activities/benthos-builder_test.go
evisdrenova: fixed

worker/pkg/workflows/datasync/activities/activities.go
```diff
@@ -627,26 +625,41 @@ func shouldProcessColumn(t *mgmtv1alpha1.JobMappingTransformer) bool {
...
+	jsFunctionStrings := strings.Join(jsFunctions, "\n")
+
+	benthosOutputString := strings.Join(benthosOutputs, "\n")
+
+	jsCode := fmt.Sprintf(`
+(() => {
+%s
+
+  const input = benthos.v0_msg_as_structured();
+  const output = { ...input_name };
```
nickzelei, line 651:
```suggestion
  const output = { ...input };
```

worker/pkg/workflows/datasync/activities/activities_test.go
```diff
@@ -502,10 +509,18 @@ func Test_buildProcessorConfigJavascriptMultiple(t *testing.T) {
...
+	benthosOutputs := []string{}
+
+	benthosOutput := constructBenthosOutput(col)
+	benthosOutput2 := constructBenthosOutput(col2)
+	benthosOutputs = append(benthosOutputs, benthosOutput, benthosOutput2)
+	jsFunction := constructJsFunction(code, col)
+	jsFunction2 := constructJsFunction(code2, col2)
+	jsFunctions = append(jsFunctions, jsFunction, jsFunction2)
+	constructedProcessor := constructBenthosJsProcessor(jsFunctions, benthosOutputs)
+	assert.Equal(t, constructedProcessor, res.Javascript.Code)
```
nickzelei, line 523: you thought it would be this easy, huh? :)

So some things about this need to change.

This test will basically never fail because the `constructedProcessor` variable here is using the same methods that `buildProcessorConfig` uses to construct `res.Javascript.Code`. So they will guaranteed to always be equivalent (because they all use the same methods!).
Now, this method would fail of course if a third column was dded and the `jsFunctions` here didn't include that. But this should be testing the output with a raw string of what you (the tester) are expecting the output to be without any of the helper functions. 

That's why in many of our other tests we run the builder function and assert it against a hardcoded benthos config string.

So these lines should really be replaced with something like:

```
assert.Equal(
  t,
  `your-expected-constructed-processor-string`,
  res.Javascript.Code,
)

File: worker/pkg/workflows/datasync/activities/benthos-builder_test.go [hunk 25]
```diff
@@ -1930,94 +1929,6 @@ func Test_ProcessorConfigEmptyJavascript(t *testing.T) {
...
-
-	res, err := bbuilder.buildBenthosSqlSourceConfigResponses(context.Background(), tableMappings, dsn, driver, sourceTableOpts)
-	assert.Nil(t, err)
-
-	out, err := yaml.Marshal(res[0].Config.Pipeline.Processors)
-	assert.NoError(t, err)
-	a, _ := yaml.Marshal(res[0].Config.Pipeline)
-	fmt.Println("res", string(a))
-
-	assert.Equal(
```
nickzelei: this is more of what I'm referring to in my above comment.
but why was this test removed?

backend/services/mgmt/v1alpha1/transformers-service/transformers.go
evisdrenova: resolved

worker/pkg/workflows/datasync/activities/activities_test.go
evisdrenova: i was hoping it would be :) updated it

worker/pkg/workflows/datasync/activities/benthos-builder_test.go
evisdrenova: resolved

worker/pkg/workflows/datasync/activities/activities_test.go
nickzelei: yea you're right we don't have any tests today that test the benthos output at an integration level.
A better approach here might be to have benthos output to a file in the temp directory that you can then read from afterwards to verify the output. That way there is no need to hijack the stdout.

backend/services/mgmt/v1alpha1/transformers-service/transformers.go
nickzelei: I was referring to the property name itself.
```suggestion
	TransformJavascript                Transformation = "transform_javascript"
```

worker/pkg/workflows/datasync/activities/activities_test.go
evisdrenova: updated

nickzelei (approved):
🚢

Issue #812, opened 2023-12-08
[NEOS-469] Make transformer name column clickable in /transformers table


nickzelei:
Might as well make all of the resources name clickable. most other tables have a uuid col and that is clickable.

<sub>From [SyncLinear.com](https://synclinear.com) | [NEOS-469](https://linear.app/nucleus-cloud/issue/NEOS-469/make-transformer-name-column-clickable-in-transformers-table)</sub>

evoxf1:
May I look into this

nickzelei:
> May I look into this

Sure! Most of the other tables in the app have the "id" column that is clickable.
The goal of this issue would be to make the transformer name column a link and it should route to the `/<account-name>/transformers/:id` page.

nickzelei:
sounds great. I've assigned it to you, please let me know if you run into any troubles.

evoxf1:
@nickzelei cool, thanks

evoxf1:
hey @nickzelei  I am having problem running the repo locally, tried composing via docker still localhost:3000 is not reachable...

evoxf1:
@nickzelei Hey Nick, I just wanna know if I am on the right track with the PR, need your feedback..

nickzelei:
> @nickzelei Hey Nick, I just wanna know if I am on the right track with the PR, need your feedback..

left some comments on your PR.

PR #1073, opened 2024-01-09
Fix Issue  #812/[NEOS-469] Make transformer name column clickable in /transformers table

evoxf1:
Fixes the issue #812 where the transformer name was not clickable.


frontend/apps/web/app/[account]/transformers/components/UserDefinedTransformersTable/data-table-row-actions.tsx
```diff
@@ -2,7 +2,7 @@
 
 import { DotsHorizontalIcon } from '@radix-ui/react-icons';
 import { Row } from '@tanstack/react-table';
-
+import Link from 'next/link';
```
Comment (nickzelei) [inline frontend/apps/web/app/[account]/transformers/components/UserDefinedTransformersTable/data-table-row-actions.tsx:5]: I don't think this import is being used anywhere in this file, is it?

frontend/apps/web/app/[account]/transformers/components/UserDefinedTransformersTable/columns.tsx
```diff
@@ -52,7 +52,14 @@ export function getUserDefinedTransformerColumns(
         return (
           <div className="flex space-x-2">
             <span className="max-w-[500px] truncate font-medium">
-              {row.original.name}
+               <div>
```
Comment (nickzelei) [inline frontend/apps/web/app/[account]/transformers/components/UserDefinedTransformersTable/columns.tsx:55]: let's just make the name clickable. no need to make the datatype or source clickable.

nickzelei (changes requested):
left a few comments, thank you for working on this!

The system transformers table also needs to be updated:
`frontend/apps/web/app/[account]/transformers/components/SystemTransformersTable/columns.tsx`

The route is slightly different, can see the `View` button that table to see, but the route will be:

`:accountName/transformers/systemTransformers/:source`

evoxf1:
> left a few comments, thank you for working on this!
> 
> The system transformers table also needs to be updated: `frontend/apps/web/app/[account]/transformers/components/SystemTransformersTable/columns.tsx`
> 
> The route is slightly different, can see the `View` button that table to see, but the route will be:
> 
> `:accountName/transformers/systemTransformers/:source`

Yes, i was just about to ask you regarding that, because I saw two folders for transformers, one client and one system. Will do the needful.. thanks @nickzelei

frontend/apps/web/app/[account]/transformers/components/UserDefinedTransformersTable/data-table-row-actions.tsx
Comment (evoxf1) [reply frontend/apps/web/app/[account]/transformers/components/UserDefinedTransformersTable/data-table-row-actions.tsx:5]: My bad. I forgot to remove that. :(

frontend/apps/web/app/[account]/transformers/components/UserDefinedTransformersTable/columns.tsx
Comment (evoxf1) [reply frontend/apps/web/app/[account]/transformers/components/UserDefinedTransformersTable/columns.tsx:55]: Actually you said in the first comment on the issue to make all the resources clickable too.

evoxf1:
> left a few comments, thank you for working on this!
> 
> The system transformers table also needs to be updated: `frontend/apps/web/app/[account]/transformers/components/SystemTransformersTable/columns.tsx`
> 
> The route is slightly different, can see the `View` button that table to see, but the route will be:
> 
> `:accountName/transformers/systemTransformers/:source`

Yes, i was just about to ask you regarding that, because I saw two folders for transformers, one client and one system. Will do the needful.. thanks @nickzelei

frontend/apps/web/app/[account]/transformers/components/UserDefinedTransformersTable/columns.tsx
Comment (nickzelei) [reply frontend/apps/web/app/[account]/transformers/components/UserDefinedTransformersTable/columns.tsx:55]: sorry for the confusion. I was referring to the resource's `name` to be clickable. We used to have the identifier in the column but it was removed to make the table look cleaner.

nickzelei (commented):
Looks good! There are some merge conflicts now, but they should hopefully be easy to resolve. I had to nest the `/app/[account]` folder under a nextjs grouping so everything is now at `/app/(mgmt)/[account]` to clean up some issues with the layout not rendering properly.

Hopefully should be an easy rebase. Let me know if you run into any issues there. 

Once the conflicts are cleaned up, we can land this PR!

frontend/apps/web/app/[account]/transformers/components/UserDefinedTransformersTable/columns.tsx
```diff
@@ -52,7 +52,14 @@ export function getUserDefinedTransformerColumns(
...
+                  <NextLink
+                    className="hover:underline"
+                    href={`/${account?.name}/transformers/${transformer.id}`}
```
Comment (nickzelei) [inline frontend/apps/web/app/[account]/transformers/components/UserDefinedTransformersTable/columns.tsx:58]: Looks like there are some build errors here with the `account` not being present.

Check out how we are adding the `accountName` to the columns component here: https://github.com/nucleuscloud/neosync/blob/main/frontend/apps/web/app/(mgmt)/%5Baccount%5D/jobs/components/DataTable/columns.tsx#L46

So the `accountName` would get added as a prop and be passed down.

nickzelei (commented):
Left another comment regarding the build errors.

evoxf1:
> Left another comment regarding the build errors.

Yes, indeed I am getting errors of sometimes packages not loading properly, the Swiggy lines wont go away even after building. Also cannot use prettier most of the times to align and indent!!

PR #1085, opened 2024-01-09
Nick/neos 622 navigating to home page results in no redirect

nickzelei:


frontend/apps/web/components/providers/account-provider.tsx
```diff
@@ -58,7 +58,7 @@ export default function AccountProvider(props: Props): ReactElement {
     const foundAccount = accountsResponse?.accounts.find(
       (a) => a.name === accountName
     );
-    if (userAccount?.id === foundAccount?.id) {
+    if (userAccount && foundAccount && userAccount?.id === foundAccount?.id) {
```
nickzelei, line 61: pesky little fucker.

hitting an edge case where `userAccount?.id === foudnAccout?.id` were both undefined which made it equal true.

Issue #1002, opened 2023-12-29
[NEOS-580] Add HPA to neosync helm charts


nickzelei:
api and worker to start

<sub>From [SyncLinear.com](https://synclinear.com) | [NEOS-580](https://linear.app/nucleus-cloud/issue/NEOS-580/add-hpa-to-neosync-helm-charts)</sub>

PR #1087, opened 2024-01-10
HPA templates to api and worker charts

brianirish:
NEOS-580
Closes https://github.com/nucleuscloud/neosync/issues/1002

**TODO:**

- [ ] Test this locally. Having trouble getting Tilt setup correctly. If someone could test this for me, I would appreciate it.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -23,6 +23,7 @@ for arg in cfg.get('to-run', []):
 config.set_enabled_resources(resources)
 
 load_dynamic('tilt/temporal/Tiltfile')
+load_dynamic('tilt/metrics-server/Tiltfile')
 load_dynamic('backend/Tiltfile')
 load_dynamic('frontend/Tiltfile')
 load_dynamic('worker/Tiltfile')
--- a/backend/charts/api/templates/deployment.yaml
+++ b/backend/charts/api/templates/deployment.yaml
@@ -94,6 +94,14 @@ spec:
             {{- toYaml . | nindent 12 }}
           {{- end }}
 
+          resources:
+            requests:
+              cpu: {{ .Values.resources.requests.cpu }}
+              memory: {{ .Values.resources.requests.memory }}
+            limits:
+              cpu: {{ .Values.resources.limits.cpu }}
+              memory: {{ .Values.resources.limits.memory }}
+
           env:
             - name: HOST_IP
               valueFrom:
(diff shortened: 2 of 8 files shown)


nickzelei:
hey @brianirish - thanks so much for the contribution!
I can test this out locally.

What issues are you having getting tilt set up locally? Happy to help out.

nickzelei:
hey Brian, I poked around on your PR a bit and had to fix a few things.

The main one is the `apiVersion` was incorrect. Just needs to be `autoscaling/v2`.

For local dev, to test I needed to add the `metrics-server` to the cluster.

That all works, however the HPA is still failing due to missing resource constraints. 
From what I read [here](https://aptakube.com/blog/how-to-fix-failedgeteesourcemetric-hpa), pods need to have a minimum cpu/ram set so that it has a baseline of utilization.

So the deployment files will need to also be updated to include a `resources.requests` section with some minimum cpu/ram set.
We probably want to have a default in the `values.yaml` much like you did for the autoscaling.

Anyways, here is a branch I created with the apiversion, metrics-server changes, but I didnt make the changes to the resources.

Here is the draft I created if you'd like to fold that into this PR: https://github.com/nucleuscloud/neosync/pull/1094
Let me know if you'd add the resources section in your PR or not!

brianirish:
That's great feedback, thanks @nickzelei! 💯 I'll hopefully be able to make progress on this later today, or this evening.

brianirish:
I've rolled in your changes @nickzelei and I'll work on the resource allocations and testing locally later today/this evening.

brianirish:
Screenshot 2024-01-11 at 9 06 53 PM

I took a screenshot of a bunch of sequential checks on CPU usage in relation to the HPA in my local kind cluster, as soon as it got deployed via `tilt up`.

The CPU defaults of 100m/300m (req/limit) seem to be more or less okay for local development and low-usage deployments. I think this makes sense for defaults, since larger deployments will want to tune for their needs anyway, and we don't want to waste resources for local/small deployments?

What do you think?

nickzelei:
> Screenshot 2024-01-11 at 9 06 53 PM
> I took a screenshot of a bunch of sequential checks on CPU usage in relation to the HPA in my local kind cluster, as soon as it got deployed via `tilt up`. I'm running an M2 Macbook Air.
> 
> The CPU defaults of 100m/300m (req/limit) seem to be more or less okay for local development and low-usage deployments. I think this makes sense for defaults, since larger deployments will want to tune for their needs anyway, and we don't want to waste resources for local/small deployments?
> 
> What do you think?

looks great! the limits seems a little low, but maybe it's okay! Would be curious to see what it spikes to during a sync or generate. it's easy enough to change during a deployment though.

Woohoo! Super happy to land this change.

nickzelei (approved):

PR #1090, opened 2024-01-10
NEOS-625:updated processors to be slices

evisdrenova:


worker/pkg/workflows/datasync/activities/activities.go
```diff
@@ -175,9 +175,15 @@ func (b *benthosBuilder) buildBenthosSqlSourceConfigResponses(
...
+		for _, pc := range processorConfig {
+			mutationValid := pc.Mutation != nil && *pc.Mutation != ""
+
+			javascriptValid := pc.Javascript != nil && pc.Javascript.Code != ""
+
+			if mutationValid || javascriptValid {
+				bc.StreamConfig.Pipeline.Processors = append(bc.StreamConfig.Pipeline.Processors, *pc)
+			}
+
 		}
```
nickzelei, line 187: this might be a little overkill. it's probably save to trust the output of `buildProcessorConfig` to ensure that the contents are correct.
This can then just become 

```
processorConfig, err := b.buildProcessorConfig(ctx, tableMappings.Mappings)
if err != nil {
  return err
}
bc.StreamConfig.Pipeline.Processors = append(bc.StreamConfig.Pipeline.Processors, *processorConfig)
```

worker/pkg/workflows/datasync/activities/activities.go
```diff
@@ -544,7 +550,7 @@ func getMysqlDsn(
 	}
 }
 
-func (b *benthosBuilder) buildProcessorConfig(ctx context.Context, cols []*mgmtv1alpha1.JobMapping) (*neosync_benthos.ProcessorConfig, error) {
+func (b *benthosBuilder) buildProcessorConfig(ctx context.Context, cols []*mgmtv1alpha1.JobMapping) ([]*neosync_benthos.ProcessorConfig, error) {
```
nickzelei, line 553: This is returning a slice now, can the method be updated to `buildProcessorConfigs`?

worker/pkg/workflows/datasync/activities/activities_test.go
```diff
@@ -360,7 +360,7 @@ func Test_buildProcessorConfigMutation(t *testing.T) {
 	})
 
 	assert.Nil(t, err)
-	assert.Equal(t, *output.Mutation, "root.id = null\nroot.name = null")
+	assert.Equal(t, *output[0].Mutation, "root.id = null\nroot.name = null")
```
nickzelei, line 363: Minor thing but you could add a new line above this that asserts the length of the output is 1, since you are expecting in your test that only a mutation config should be added.

```
assert.Len(t, output, 1)
assert.Equal(t, *output[0].Mutation, "root.id = null\nroot.name = null")
```

Then, for any test where you are expecting a mutation _and_ js processor, you can assert that the length should be 2. and so on. Although i dont see a test that handles the case where both are present. Would be great to have one.

worker/pkg/workflows/datasync/activities/benthos-builder.go
```diff
@@ -595,8 +595,18 @@ func (b *benthosBuilder) buildBenthosGenerateSourceConfigResponses(
 		if err != nil {
 			return nil, err
 		}
-		if mapping.Mutation == nil {
-			return nil, errors.New("unable to generate config mapping for table") // workshop this more
+
+		var mutationMapping string
```
nickzelei, line 599: So the `buildProcessorConfig` builds both a `mutation` and `javascript` config right?

Maybe that method could be further broken down internally to look something like:

```
func buildProcessorConfig(...) {
   mutation := buildMutationConfig(...)

   js := buildJsConfig(...)
   
  // construct slice
  return processorConfigs
}
```

Then, for the context of this method, instead of calling `buildProcessorConfig` and having to iterate over the output, you could then just call:
```
mapping := b.buildMutationConfig(ctx, tableMapping.Mappings)
```

worker/pkg/workflows/datasync/activities/activities.go
evisdrenova: udpated

worker/pkg/workflows/datasync/activities/activities.go
evisdrenova: Removed extra checks. Since processorConfigs returns a slice so we have to range over it and append each object to the processors.

worker/pkg/workflows/datasync/activities/activities_test.go
evisdrenova: Updated. Test that handles both is in the benthos-builder_test.go called `Test_ProcessorConfigMutationAndJavascript`

worker/pkg/workflows/datasync/activities/benthos-builder.go
evisdrenova: refactored and updated

nickzelei:
let's chat in the morning about your latest changes. Need to make a few more revisions and it will be easier to go through it in person.

worker/pkg/workflows/datasync/activities/activities.go
```diff
@@ -544,57 +543,66 @@ func getMysqlDsn(
 	}
 }
 
-func (b *benthosBuilder) buildProcessorConfig(ctx context.Context, cols []*mgmtv1alpha1.JobMapping) (*neosync_benthos.ProcessorConfig, error) {
+func (b *benthosBuilder) buildProcessorConfigs(ctx context.Context, cols []*mgmtv1alpha1.JobMapping) ([]*neosync_benthos.ProcessorConfig, error) {
+	var jsFunctions, benthosOutputs []string
+	var err error
+
+	jsFunctions, benthosOutputs = b.extractJsFunctionsAndOutputs(cols, jsFunctions, benthosOutputs)
```
nickzelei, line 550: Why is this not just:
```
jsFunctions, benthosOutputs = b.extractJsFunctionsAndOutputs(cols)
```

Or even better:

```
jsCode := b.buildJsCodeConfig(cols)
```

This reduces this function to:

```
mutation, err := b.buildMutationConfig(ctx, cols)
if err != nil {
  return err
}
jsCode, err := b.buildJsCodeConfig(ctx, cols)
if err != nil {
  return err
}

processorConfigs := []*neosync_benthos.ProcessorConfig{}
if len(mutation) > 0 {
  processorConfigs = append(processorConfigs, &neosync_benthos.ProcessorConfig{Mutation: &mutation})
}
if len(jsCode) > 0 {
  processorConfigs = append(processorConfigs, &neosync_benthos.ProcessorConfig{Javascript: &JavascriptConfig{Code: jsCode}})
}
return processorConfigs
```

evisdrenova: @nickzelei yeah i suppose can we do that too :) updated

nickzelei (approved):
:shipit:

PR #1177, opened 2024-01-23
Adds worker api key type

nickzelei:
* Adds new worker api key token type
* Updates api key middleware to support worker api keys
* api key mw supports a whitelist of api keys and procedures
* Updated relevant services to support worker api keys for the verify in account check

### DIFF
--- a/backend/internal/cmds/mgmt/serve/connect/cmd.go
+++ b/backend/internal/cmds/mgmt/serve/connect/cmd.go
@@ -165,7 +165,7 @@ func serve(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
-		apikeyClient := auth_apikey.New(db.Q, db.Db)
+		apikeyClient := auth_apikey.New(db.Q, db.Db, getAllowedWorkerApiKeys(getIsNeosyncCloud()), []string{"TODO"})
 		stdAuthInterceptors = append(
 			stdAuthInterceptors,
 			auth_interceptor.NewInterceptor(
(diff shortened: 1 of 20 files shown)


backend/internal/cmds/mgmt/serve/connect/cmd.go
```diff
@@ -165,7 +165,11 @@ func serve(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
-		apikeyClient := auth_apikey.New(db.Q, db.Db)
+		apikeyClient := auth_apikey.New(db.Q, db.Db, getAllowedWorkerApiKeys(getIsNeosyncCloud()), []string{
+			mgmtv1alpha1connect.JobServiceGetJobProcedure,
+			mgmtv1alpha1connect.ConnectionServiceGetConnectionProcedure,
+			mgmtv1alpha1connect.TransformersServiceGetUserDefinedTransformerByIdProcedure,
```
nickzelei, line 171: when providing a worker api key, these are the only methods that they are valid for.

backend/internal/cmds/mgmt/serve/connect/cmd.go
```diff
@@ -584,3 +588,10 @@ func getAuthApiClientSecret() string {
 func getIsNeosyncCloud() bool {
 	return viper.GetBool("NEOSYNC_CLOUD")
 }
+
+func getAllowedWorkerApiKeys(isNeosyncCloud bool) []string {
+	if isNeosyncCloud {
```
nickzelei, line 593: will only load api keys if in neosync cloud. why? well, that's it's only main usecase for right now.

alishakawaguchi (approved):
🥇

PR #1183, opened 2024-01-24
NEOS-673:add in column filtering for headers and button

evisdrenova:


nickzelei (changes requested):
this needs some work due to perf implications. will be taking this over, but we're fixing a few other things on this form first.

PR #1211, opened 2024-01-29
Nick/neos 621 Makes retrieving Job Runs more performant

nickzelei:


frontend/apps/web/app/(mgmt)/[account]/runs/[id]/components/JobRunActivityTable/data-table.tsx
```diff
@@ -120,7 +120,7 @@ export function DataTable({ columns, data, isError }: DataTableProps) {
                   colSpan={columns.length}
                   className="h-24 text-center"
                 >
-                  No runs in the last 24 hours.
+                  No active runs found
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/runs/[id]/components/JobRunActivityTable/data-table.tsx:123]: we can't hard code the window because it's based on when temporal has been configured to archive runs.

alishakawaguchi (approved):
👍

PR #1232, opened 2024-01-30
move oidc signout to get instead of post

nickzelei:


66244484:
I like to program process sounds nice and sucre ..

PR #1233, opened 2024-01-30
NEOS-669: created transform character scramble transformer in worker

evisdrenova:


worker/internal/benthos/transformers/transform_character_substitution.go
```diff
@@ -0,0 +1,100 @@
...
+		'0': '1', '1': '2', '2': '3', '3': '4', '4': '5', '5': '6', '6': '7', '7': '8', '8': '9', '9': '0',
+	}
+
+	specialCharMap := map[rune]rune{
+		'!': '@', '@': '#', '#': '$', '$': '%', '%': '^', '^': '&', '&': '*',
+		'*': '(', '(': ')', ')': '-', '-': '+', '+': '=', '=': '_',
+		'_': '[', '[': ']', ']': '{', '{': '}', '}': '|', '|': '\\',
+		'\\': ':', ':': ';', ';': '\'', '\'': '"', '"': '<', '<': '>',
+		'>': ',', ',': '.', '.': '/', '/': '?', '?': '!',
+	}
```
nickzelei, line 75: the way this is defined now these get instantiated every time. if you move them to the module level then they get instantiated once and this method will be _that_ much more efficient.

worker/internal/benthos/transformers/transform_character_substitution.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if unicode.IsLetter(r) {
```
nickzelei, line 77: talked offline, the substitution should be random instead of the same every time.

worker/internal/benthos/transformers/utils/string_utils.go
```diff
@@ -121,3 +122,23 @@ func IsValidUsername(username string) bool {
...
+}
+
+func IsAllowedSpecialChar(r rune) bool {
+	allowedSpecialChars := "!@#$%^&*()-+=_[]{}|\\ ;\"<>,./?"
+	for _, allowedChar := range allowedSpecialChars {
+		if r == allowedChar {
+			return true
+		}
+	}
+	return false
```
nickzelei, line 143: this can be made much faster over time if you turn this into a set.
```
var allowedSpecialChars = map[rune]any{"!": struct{}{}, ...}
func IsAllowedSpecialChar(r rune) bool {
  _, ok := allowedSpecialChars[r]
 return ok
}
```

nickzelei (commented):
left a few comments inline

agreed on renaming to something more specific like scramble

worker/internal/benthos/transformers/transform_character_substitution.go
evisdrenova: updated to remove map and instead just use strings

worker/internal/benthos/transformers/transform_character_substitution.go
evisdrenova: updated

worker/internal/benthos/transformers/utils/string_utils.go
evisdrenova: updated this by creating a map of the allowed special characters and using to check if a character is is a specal character or not

worker/internal/benthos/transformers/utils/string_utils.go
```diff
@@ -134,11 +134,12 @@ func IsValidChar(s string) bool {
 }
 
 func IsAllowedSpecialChar(r rune) bool {
-	allowedSpecialChars := "!@#$%^&*()-+=_[]{}|\\ ;\"<>,./?"
-	for _, allowedChar := range allowedSpecialChars {
-		if r == allowedChar {
-			return true
-		}
+	allowedSpecialChars := map[rune]struct{}{
```
nickzelei, line 137: It's beneficial to move stuff like this outside of the function so that it doesn't get re-defined every time the function is invoked. 

If it's moved to be defined at the module level, then it's defined and allocated once by the process, and then never again. Where as the way it currently is, the process has to define this map, check if the rune is in it, then throw it away. 

Should look something like this:

```
var (
  allowedSpecialChars = map[rune]struct{}{...}
)

func IsAllowedSpecialChar(r rune) bool {
  _, ok := allowedSpecialChars[r]
  return ok
}
```

worker/internal/benthos/transformers/transform_character_substitution.go
```diff
@@ -48,52 +49,57 @@ Substituted: Ifmmp Xpsme 234@%^
...
 }
 
-func substituteChar(r rune) rune {
-	letterMap := map[rune]rune{
-		'a': 'b', 'b': 'c', 'c': 'd', 'd': 'e', 'e': 'f', 'f': 'g', 'g': 'h', 'h': 'i', 'i': 'j',
-		'j': 'k', 'k': 'l', 'l': 'm', 'm': 'n', 'n': 'o', 'o': 'p', 'p': 'q', 'q': 'r', 'r': 's',
-		's': 't', 't': 'u', 'u': 'v', 'v': 'w', 'w': 'x', 'x': 'y', 'y': 'z', 'z': 'a',
-	}
+func ScrambleChar(r rune) rune {
+	letterList := "abcdefghijklmnopqrstuvwxyz"
```
nickzelei, line 60: let's get these lists moved outside of the function in a module-level const block so they are only defined a single time.

worker/internal/benthos/transformers/transform_character_substitution.go
```diff
@@ -48,52 +49,57 @@ Substituted: Ifmmp Xpsme 234@%^
...
 
-	if sub, ok := specialCharMap[r]; ok {
-		return sub
+	randInd, err := transformer_utils.GenerateRandomInt64InValueRange(0, int64(len(specialCharList))-1)
+	if err != nil {
+		return r
+	}
+
+	if transformer_utils.IsAllowedSpecialChar(r) {
+		if unicode.IsSpace(r) {
```
nickzelei, line 97: little Q here, but should this check happen before anything else? Seems like we alway want to return the space in a string, where as right now we only check if it's a space if it's an allowed special char.

could the logic flow of this function be:

```
if unicode.IsLetter(r) {}
else if unicode.IsSpace(r) {}
else if unicode.IsNumber(r) {}
else if IsAllowedSpecialChar(r) {}
```

worker/internal/benthos/transformers/transform_character_substitution.go
```diff
@@ -48,52 +49,57 @@ Substituted: Ifmmp Xpsme 234@%^
...
+
+	specialCharList := "!@#$%^&*()-+=_ []{}|\\ ;\"<>,./?"
 
-	specialCharMap := map[rune]rune{
-		'!': '@', '@': '#', '#': '$', '$': '%', '%': '^', '^': '&', '&': '*',
-		'*': '(', '(': ')', ')': '-', '-': '+', '+': '=', '=': '_',
-		'_': '[', '[': ']', ']': '{', '{': '}', '}': '|', '|': '\\',
-		'\\': ':', ':': ';', ';': '\'', '\'': '"', '"': '<', '<': '>',
-		'>': ',', ',': '.', '.': '/', '/': '?', '?': '!',
+	randStringListInd, err := transformer_utils.GenerateRandomInt64InValueRange(0, int64(len(letterList))-1)
```
nickzelei, line 66: This is def a micro optimization, but because we know the lists are a constant, we could pre-compute the length and store that as a module-level constant as well, so every function invocation doesn't have to re-calculate the length of the letterlist (or any of the other lists, where as you do right now because it's a string.

nickzelei (commented):
overall lgtm, left a few more comments for some optimizations.

worker/internal/benthos/transformers/utils/string_utils.go
evisdrenova: makes sense, updated

worker/internal/benthos/transformers/transform_character_substitution.go
evisdrenova: updated

worker/internal/benthos/transformers/transform_character_substitution.go
evisdrenova: refactored to htis

PR #1243, opened 2024-01-31
Helm chart fixes

brianirish:
- [all] Made `shutdownTimeoutSeconds` an optional field
- [all] Made `terminationGracePeriod` an optional field
- [api] Made `neosyncCloud.workerApiKeys` an optional field
- [api] Fixed the `namespace` for the role and role-binding so it's based on the chart release
- [app] Made `neosyncApi.url` an optional field

### DIFF
--- a/backend/charts/api/templates/api-env-vars.yaml
+++ b/backend/charts/api/templates/api-env-vars.yaml
@@ -26,7 +26,9 @@ stringData:
     NUCLEUS_ENV: {{ .Values.nucleusEnv }}
     {{- end }}
 
+    {{- if .Values.shutdownTimeoutSeconds }}
     SHUTDOWN_TIMEOUT_SECONDS: {{ .Values.shutdownTimeoutSeconds | quote }}
+    {{- end }}
 
     {{- if and .Values.auth .Values.auth.enabled }}
     AUTH_ENABLED: {{ .Values.auth.enabled | default "false" | quote }}
@@ -48,15 +50,15 @@ stringData:
     AUTH_CLIENTID_SECRET: {{ .Values.auth.clientMap | toJson | quote }}
     {{- end }}
 
-   {{- if and .Values.auth .Values.auth.cliClientId }}
+    {{- if and .Values.auth .Values.auth.cliClientId }}
     AUTH_CLI_CLIENT_ID: {{ .Values.auth.cliClientId }}
     {{- end }}
 
-   {{- if and .Values.auth .Values.auth.cliAudience }}
+    {{- if and .Values.auth .Values.auth.cliAudience }}
     AUTH_CLI_AUDIENCE: {{ .Values.auth.cliAudience }}
     {{- end }}
 
-   {{- if and .Values.auth .Values.auth.signatureAlgorithm }}
+    {{- if and .Values.auth .Values.auth.signatureAlgorithm }}
     AUTH_SIGNATURE_ALGORITHM: {{ .Values.auth.signatureAlgorithm }}
     {{- end }}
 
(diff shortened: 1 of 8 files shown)


backend/charts/api/templates/api-env-vars.yaml
```diff
@@ -26,7 +26,7 @@ stringData:
     NUCLEUS_ENV: {{ .Values.nucleusEnv }}
     {{- end }}
 
-    SHUTDOWN_TIMEOUT_SECONDS: {{ .Values.shutdownTimeoutSeconds | quote }}
+    SHUTDOWN_TIMEOUT_SECONDS: {{ .Values.shutdownTimeoutSeconds | default "300" | quote }}
```
nickzelei, line 29: One other way we could go with this is to only add the shutdown timeout to the deployment if the value is present. That way the deployment will just defer to the Kubernetes default instead of our hardcoded one.
wdyt?

backend/charts/api/templates/api-env-vars.yaml
```diff
@@ -101,7 +101,9 @@ stringData:
     {{- end }}
 
     NEOSYNC_CLOUD: {{ .Values.neosyncCloud.enabled | default "false" | quote }}
+    {{- if .Values.neosyncCloud.enabled }}
     NEOSYNC_CLOUD_ALLOWED_WORKER_API_KEYS: {{ join "," .Values.neosyncCloud.workerApiKeys }}
+    {{- end }}
```
nickzelei, line 106: This is fine, for what it's worth, this value is only checked in the code if the above is true. Definitely not clear from looking at the helm chart though.

```
func getAllowedWorkerApiKeys(isNeosyncCloud bool) []string {
	if isNeosyncCloud {
		return viper.GetStringSlice("NEOSYNC_CLOUD_ALLOWED_WORKER_API_KEYS")
	}
	return []string{}
}
```

backend/charts/api/templates/role-binding.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  namespace: neosync
+  namespace: {{ .Release.Namespace }}
```
nickzelei, line 4: 👍

backend/charts/api/templates/role.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  namespace: neosync
+  namespace: {{ .Release.Namespace }}
```
nickzelei, line 4: 👍

nickzelei (commented):
thanks for making these improvements, left one comment regarding the shut down timer

backend/charts/api/templates/api-env-vars.yaml
brianirish: I think you're right, that's a better solution! I'll make the change shortly. Thanks!

backend/charts/api/templates/api-env-vars.yaml
brianirish: Oh, I see! Thanks for clarifying. The only reason I made this change was because empty strings aren't allowed in Kubernetes secret stringData, so a value must be present. I was getting an error omitting `workerApiKeys` from my values.yaml file.

nickzelei: ah! totally makes sense.

nickzelei (commented):
last thing @brianirish 

I think for the grace period timeout, we probably want to wrap the setting in the deployment in an if so that it doesnt get set to an empty string? K8s might handle it ok, but maybe it's worth being safe

This is what it looks like now, so if it's not provided it might act funky.
```
terminationGracePeriod: {{ .Values.terminationGracePeriod }}
```

brianirish:
@nickzelei You got it. I've committed that change. 👍

nickzelei (approved):
🚢

PR #1282, opened 2024-02-07
Refactor schema page table

alishakawaguchi:


frontend/apps/web/components/jobs/SchemaTable/schema-table.tsx
```diff
@@ -50,6 +51,11 @@ export function SchemaTable(props: Props): ReactElement {
     };
   });
 
+  const columns = useMemo(
+    () => getSchemaColumns({ transformers: mergedTransformers }),
+    [mergedTransformers] // Use mergedTransformers instead of transformers
```
nickzelei, line 56: I don't think this is actually doing anything because every render the mergedTransformers is created and thus, a new object. So every render will trigger this useMemo.
Can leave this as-is for now though or take it out and can figure out the transformer optimization later.

frontend/package.json
```diff
@@ -18,5 +18,8 @@
 	"workspaces": [
 		"apps/*",
 		"packages/*"
-	]
+	],
+	"dependencies": {
+		"@faker-js/faker": "^8.4.0"
+	}
```
nickzelei, line 24: do we need this here?

frontend/apps/web/components/jobs/SchemaTable/schema-table.tsx
evisdrenova: i ended up just putting the merged transformers into a useMemo so now it only re-renders if there's a change in one of the transformers (which should really only ever be the customTransformers and even then it's unlikely for that to change while someone is on the schema page).

evisdrenova: actually im just going to put it into a hook

PR #1307, opened 2024-02-09
NEOS-752: support for email transformer exclusion list

evisdrenova:


worker/internal/benthos/transformers/transform_email.go
```diff
@@ -44,9 +45,29 @@ func init() {
...
 		}
 
+		eL, err := args.Get("exclusion_list")
+		if err != nil {
+			return nil, err
+		}
+
+		excl, ok := eL.([]any)
+		if !ok {
+			return nil, fmt.Errorf("error")
```
nickzelei, line 55: just so this doesn't get lost: this error should be updated. 
Should do something like this:

```
fmt.Errorf("Expected []any, got: %T", excl)
```

worker/internal/benthos/transformers/transform_email.go
```diff
@@ -44,9 +45,29 @@ func init() {
...
+		}
+
+		var new []string
+
+		for _, a := range excl {
+			new = append(new, a.(string))
```
nickzelei, line 61: need to check the type here.
```
strval, ok := a.(string)
if !ok {
  return nil, fmt.Errorf("Expected string, got :%T", a)
}
```

nickzelei (commented):
left a few comments we talked about offline. overall would be great to have some better variable names for readability too

worker/internal/benthos/transformers/transform_email.go
evisdrenova: updated

worker/internal/benthos/transformers/transform_email.go
evisdrenova: udpated

PR #1339, opened 2024-02-12
fixed street address transformer and POC for seeds

evisdrenova:


worker/internal/benthos/transformers/generate_street_address.go
```diff
@@ -20,16 +20,28 @@ type Address struct {
 }
 
 func init() {
-	spec := bloblang.NewPluginSpec().Param(bloblang.NewInt64Param("max_length"))
+	spec := bloblang.NewPluginSpec().Param(bloblang.NewInt64Param("max_length")).Param(bloblang.NewInt64Param("seed"))
```
nickzelei, line 23: This needs to be marked as optional so that it can be omitted from the bloblang interface.
```suggestion
	spec := bloblang.NewPluginSpec().Param(bloblang.NewInt64Param("max_length")).Param(bloblang.NewInt64Param("seed").Optional())
```

worker/pkg/workflows/datasync/activities/activities.go
```diff
@@ -727,6 +728,11 @@ func computeMutationFunction(col *mgmtv1alpha1.JobMapping, colInfo *dbschemas_ut
 		maxLen = *colInfo.CharacterMaximumLength
 	}
 
+	// Use the global RNG to seed the transformer RNGs
+	seed := rand.Int()
+
+	seedValue := rand.New(rand.NewSource(int64(seed)))
```
nickzelei, line 734: can we just leave this out for now and let the transformer internals generate this in the init method?

worker/internal/benthos/transformers/generate_street_address_test.go
```diff
@@ -2,47 +2,64 @@ package transformers
...
+	assert.NoError(t, err)
+
+	assert.IsType(t, Address{}.Address1, res, "The returned street address should be a string")
+	assert.LessOrEqual(t, int64(len(res.(string))), maxLength, fmt.Sprintf("The street should be less than or equal to the max length. This is the error street address:%s", res))
+}
+
+func Test_StreetAddressTransformerNoSeed(t *testing.T) {
+
+	var noSeed *int64
+	mapping := fmt.Sprintf(`root = generate_street_address(max_length:%d: seed:%d)`, maxLength, noSeed)
```
nickzelei, line 56: if there is no seed, can we just omit the seed parameter entirely from this test?
```suggestion
	mapping := fmt.Sprintf(`root = generate_street_address(max_length:%d)`, maxLength)
```

worker/internal/benthos/transformers/generate_street_address.go
```diff
@@ -20,16 +20,28 @@ type Address struct {
...
 		}
 
+		seed, err := args.GetOptionalInt64("seed")
+		if err != nil {
+			return nil, err
+		}
+
+		var randomizer *rand.Rand
+		if seed == nil {
+			randomizer = rand.New(rand.NewSource(int64(rand.Int())))
```
nickzelei, line 38: I think you can just use `rand.Int63` and have it return an int64 for you automatically.
```suggestion
			randomizer = rand.New(rand.NewSource(rand.Int63()))
```

worker/internal/benthos/transformers/generate_street_address.go
evisdrenova: updated

worker/internal/benthos/transformers/generate_street_address.go
evisdrenova: yup you're right!

worker/pkg/workflows/datasync/activities/activities.go
evisdrenova: Yeah we can do that, then we just won't pass it in. Updated.

worker/internal/benthos/transformers/generate_street_address_test.go
evisdrenova: yup

worker/internal/benthos/transformers/generate_street_address.go
```diff
@@ -20,16 +20,27 @@ type Address struct {
...
+		seed, err := args.GetOptionalInt64("seed")
+		if err != nil {
+			return nil, err
+		}
+		var randomizer *rand.Rand
+		if seed == nil {
+			randomizer = rand.New(rand.NewSource(rand.Int63()))
+		} else {
+			randomizer = rand.New(rand.NewSource(*seed))
+		}
```
nickzelei, line 40: Separately, we'll probably want to pull this into a func so that other transformers can more easily instantiate a randomizer.

```
seed, err := args.GetOptionalInt64("seed")
if err != nil {
  return nil, err
}
randomizer := getRandomizer(seed)
```

nickzelei (commented):
I think this overall looks good. Looks like there will be some subset of transformers that will require updating in order to facilitate this randomizer update.

Up to you if you'd rather separate the seed POC into a separate PR and get the City fix merged in first.

worker/internal/benthos/transformers/generate_street_address.go
evisdrenova: yeah im gonna fix the street address one, and then split this out into a separte PR and then come back to it later this week

PR #1468, opened 2024-03-01
NEOS-819: ability to clone connections from the connection form

evisdrenova:


frontend/apps/web/app/(mgmt)/[account]/connections/[id]/page.tsx
```diff
@@ -49,15 +54,19 @@ export default function ConnectionPage({ params }: PageProps) {
         variant: 'destructive',
       }),
     extraPageHeading: (
-      <div>
+      <div className="flex flex-row items-center gap-4">
+        <CloneConnectionButton
+          connectionType={data?.connection?.connectionConfig?.config.case}
+          id={data?.connection?.id}
+        />
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/connections/[id]/page.tsx:61]: minor thing here but let's not render this button unless these data types are present.

frontend/apps/web/app/(mgmt)/[account]/connections/[id]/page.tsx
```diff
@@ -69,3 +78,39 @@ export default function ConnectionPage({ params }: PageProps) {
     </OverviewContainer>
   );
 }
+
+interface CloneConnectionProps {
+  connectionType?: string;
+  id?: string;
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/connections/[id]/page.tsx:84]: these should be required if they are needed for the method to function.

marketing/app/pricing/DotBackground copy.tsx
```diff
@@ -0,0 +1,15 @@
+import { ReactNode } from 'react';
```
nickzelei, line 1: what is this?

frontend/apps/web/app/(mgmt)/[account]/new/connection/postgres/PostgresForm.tsx
```diff
@@ -130,6 +136,66 @@ export default function PostgresForm() {
...
+            });
+          }
+        } catch (error) {
+          console.error('Failed to fetch connection data:', error);
+        }
+      }
+    };
+
+    fetchData();
+  }, [sourceConnId, account, form]);
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/new/connection/postgres/PostgresForm.tsx:193]: you only want this useEffect to run once, so it should not have any dependencies, just an empty `[]`.

Comment (nickzelei) [reply frontend/apps/web/app/(mgmt)/[account]/new/connection/postgres/PostgresForm.tsx:193]: If you're worried about the account id not being present, then there needs to be a trigger to ensure that this fetch only happens a single time and doesn't re-run and reset the form on the user.

Also, it's generally a bad idea to use objects as a dependency unless you're 100% sure they are memoized. It's generally better to key off one of the properties that you're using instead as it's a stable value. For example you're using only the `account.id` here. The dependency should instead be `account?.id` instead of `account`.

frontend/apps/web/app/(mgmt)/[account]/new/connection/postgres/PostgresForm.tsx
```diff
@@ -130,6 +136,66 @@ export default function PostgresForm() {
     }
   }
 
+  /* we call the underlying useGetConnection API directly since we can't call
+the hook in the useEffect conditionally. 
+*/
+  useEffect(() => {
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/new/connection/postgres/PostgresForm.tsx:142]: I hate useEffects, but overall I think this makes sense here given how we want to do this.

I'd like to see this method cleaned up a bit though prior to merge to improve its readability.

Generally when we make requests we pull them out into a function to abstract away all of the data handling. tons of examples of that around the app. Should clean this up considerably.

nickzelei (changes requested):
What's going on here with this submodule?
image

Left a few comments (i know this is still a WIP, but wanted to call them out).

Overall I think this direction makes sense given the security implications. I was thinking it could be nice to store in session state or in query params, but idk if we really want to store sensitive data like that in that medium, so pulling from the backend makes the most sense I think.

frontend/apps/web/app/(mgmt)/[account]/new/connection/postgres/PostgresForm.tsx
```diff
@@ -130,6 +136,66 @@ export default function PostgresForm() {
...
+                user: pgConfig.user,
+                pass: pgConfig.pass,
+                port: pgConfig.port,
+                sslMode: pgConfig.sslMode,
+              },
+            });
+          }
+        } catch (error) {
+          console.error('Failed to fetch connection data:', error);
+        }
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/new/connection/postgres/PostgresForm.tsx:188]: You need to make sure that after all else, the isLoading state is set to false, otherwise the form is unusable. Add a `finally{}` block here to ensure that `setIsLoading` is always set to false. Then it does not need to be present in the try or catch blocks.

frontend/apps/web/app/(mgmt)/[account]/connections/[id]/page.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/connections/[id]/page.tsx:84]: yeah these are required for the routing to work correctly.

marketing/app/pricing/DotBackground copy.tsx
evisdrenova: removed this, doesn't belong here

frontend/apps/web/app/(mgmt)/[account]/new/connection/postgres/PostgresForm.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/new/connection/postgres/PostgresForm.tsx:142]: Yeah totally agree, cleaned it up a bit.

frontend/apps/web/app/(mgmt)/[account]/new/connection/postgres/PostgresForm.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/new/connection/postgres/PostgresForm.tsx:193]: makes sense, updated it so that only the account.Id is in the dependency array and the page waits for the `useAccount` hook to return before rendering the page

frontend/apps/web/app/(mgmt)/[account]/connections/components/ConnectionsTable/columns.tsx
```diff
@@ -67,7 +67,9 @@ export function getColumns(
         return (
           <div className="flex space-x-2">
             <span className="max-w-[500px] truncate font-medium">
-              {getCategory(row.original.connectionConfig)}
+              {getCategory(row.original.connectionConfig) == 'Mysql'
+                ? 'MySQL'
+                : getCategory(row.original.connectionConfig)}
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/connections/components/ConnectionsTable/columns.tsx:72]: why not just have the `getCategory` function return `MySQL`? That is the entire point of that method to return the value to be displayed in the table :D

frontend/apps/web/app/(mgmt)/[account]/new/connection/aws-s3/AwsS3Form.tsx
```diff
@@ -69,6 +75,65 @@ export default function AwsS3Form() {
...
+        setIsLoading(true);
+        try {
+          const connData = await GetConnectionCloneValues(
+            account.id,
+            sourceConnId
+          );
+
+          if (connData) {
+            const s3Config = connData.connection?.connectionConfig?.config
+              .value as AwsS3ConnectionConfig;
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/new/connection/aws-s3/AwsS3Form.tsx:93]: No guarantee that the connection id that is provided in the url is actually an S3 connection. So you need to actually check to make sure this is valid here and not type cast it because otherwise it might be undefined.
Use the typeguards to check that it is actually an s3 connection.

Same should be done for the mysql/postgres forms too.

nickzelei (commented):
overall looks good, left another comment around validating the incoming source id

frontend/apps/web/app/(mgmt)/[account]/connections/components/ConnectionsTable/columns.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/connections/components/ConnectionsTable/columns.tsx:72]: ah i missed that it was a func we wrote, thought it was from tanstack, that's much easier!

frontend/apps/web/app/(mgmt)/[account]/new/connection/aws-s3/AwsS3Form.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/new/connection/aws-s3/AwsS3Form.tsx]: updated

nickzelei (approved):

PR #1475, opened 2024-03-02
wip on updating benethos input raw

nickzelei:


worker/pkg/workflows/datasync/activities/gen-benthos-configs/benthos-builder.go
```diff
@@ -962,7 +973,13 @@ func createSqlUpdateBenthosConfig(
 			out := buildMysqlOutputQueryAndArgs(newResp, tm, tableKey, colSourceMap)
 			output = out
 		}
-		newResp.Columns = output.Columns
+		if newResp.Config.Input.SqlSelect != nil {
+			newResp.Config.Input.SqlSelect.Where = insertConfig.Config.Input.SqlSelect.Where // keep the where clause the same as insert
+			newResp.Columns = output.Columns
+		} else if newResp.Config.Input.PooledSqlRaw != nil {
+			// buildSelectQuery(driver, tableKey, "", output.Columns)
+			newResp.Config.Input.PooledSqlRaw.Query = insertConfig.Config.Input.PooledSqlRaw.Query // todo: should be updated to include the column subset
```
nickzelei: todo: have alisha look at this

worker/pkg/workflows/datasync/activities/gen-benthos-configs/benthos-builder.go
```diff
@@ -962,7 +973,12 @@ func createSqlUpdateBenthosConfig(
 			out := buildMysqlOutputQueryAndArgs(newResp, tm, tableKey, colSourceMap)
 			output = out
 		}
-		newResp.Columns = output.Columns
+		if newResp.Config.Input.SqlSelect != nil {
+			newResp.Config.Input.SqlSelect.Where = insertConfig.Config.Input.SqlSelect.Where // keep the where clause the same as insert
+			newResp.Columns = output.Columns
+		} else if newResp.Config.Input.PooledSqlRaw != nil {
+			newResp.Config.Input.PooledSqlRaw.Query = insertConfig.Config.Input.PooledSqlRaw.Query // todo: should this be updated to include the column subset?
```
nickzelei, line 980: @alishakawaguchi - is this okay here? 
For the sql select looks like we provide an updated subset of columns, but the same WHERE statement.
However, for the raw we just have the fully built sql statement, it wont be as easy to hot swap out portions of the query.
Is it imperative that we provide the updated column set here? Trying to understand the reasoning behind it. Maybe we can chat about it offline in office.

worker/pkg/workflows/datasync/activities/gen-benthos-configs/query-builder.go
```diff
@@ -0,0 +1,35 @@
...
+	selectColumns := make([]any, len(columns))
+	for i, col := range columns {
+		selectColumns[i] = col
+	}
+	query := builder.From(sqltable).Select(selectColumns...)
+
+	if whereClause != nil && *whereClause != "" {
+		query = query.Where(goqu.L(*whereClause))
+	}
+	sql, _, err := query.Prepared(true).ToSQL()
```
nickzelei: we will eventually want to return the parameters here. however, the user provides the where statment today so we'd need to run logic over it to parameterize it before we pass it in to Goqu, as goqu only parameterizes if you provide a Where sturct instead of the already stringified version.

PR #1622, opened 2024-03-21
NEOS-910: updated test connection query to check that we can see schemas and ta…

evisdrenova:
Adding in support in postgres and mysql to try to retrieve schema and table data to ensure that we have access to that data. By default, every user in postgres can read from the information_schema, so a company would have to explicitly turn that off. So if we can't see the schema, there are 3 possible reasons:

1. The schema(s) "exists" but is empty i.e. no tables
2. The user that neosync is using to connect does not have access to the schema(s)
3. There is something wrong with the database where we cannot connect at all. 

If there are schemas/tables, the response is something like:
```
{
    "statistics": [
        {
            "schema": "neosync_source",
            "table_count": "6"
        },
        {
            "schema": "public",
            "table_count": "1"
        }
    ],
    "is_connected": true
}
```

Then on the front-end we can render the schemas and table counts that we see. We can say somethign like: "here's what we found: public  (4 tables), neosync_source (3 tables), etc..

There is a possibility that someone has a ton of schemas which would expand the size of the success alert. We could alternatively, just return the number of schemas we see to limit the dimensionality of the data. So we say something like "We found 4 schemas!".


backend/services/mgmt/v1alpha1/connection-service/connection.go
```diff
@@ -39,17 +39,99 @@ func (s *Service) CheckConnectionConfig(
 
 	cctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
-	err = db.PingContext(cctx)
+	// err = db.PingContext(cctx)
```
nickzelei, line 42: Probably still want to have this as a first line sanity check

backend/services/mgmt/v1alpha1/connection-service/connection.go
```diff
@@ -39,17 +39,99 @@ func (s *Service) CheckConnectionConfig(
...
+	tx, err := db.BeginTx(cctx, &sql.TxOptions{ReadOnly: true})
```
nickzelei, line 43: why does this need to happen inside a transaction?

backend/services/mgmt/v1alpha1/connection-service/connection.go
```diff
@@ -39,17 +39,99 @@ func (s *Service) CheckConnectionConfig(
...
+	pgQuery := `
+	SELECT 
+		nspname AS schema_name, 
+		COUNT(*) AS number_of_tables
+	FROM 
+		pg_catalog.pg_tables
+		JOIN pg_catalog.pg_namespace ON pg_namespace.nspname = pg_tables.schemaname
+	WHERE 
+		pg_namespace.oid NOT IN 
+			(SELECT oid FROM pg_catalog.pg_namespace WHERE nspname = 'pg_catalog' OR nspname = 'information_schema')
```
nickzelei, line 60: I think this query can be reduced down to a single table.
```
SELECT 
  schemaname, 
  COUNT(*) AS number_of_tables
FROM 
  pg_catalog.pg_tables
WHERE
  schemaname NOT IN ('pg_catalog', 'information_schema')
GROUP BY schemaname;
```

nickzelei: Also I don't know if the order by is really necessary unless you are using it to display more nicely on the frontend.

backend/services/mgmt/v1alpha1/connection-service/connection.go
```diff
@@ -39,17 +39,99 @@ func (s *Service) CheckConnectionConfig(
...
+			IsConnected:     false,
+			ConnectionError: &errorMsg,
+			Statistics:      nil,
+		}), nil
+	}
+	defer rows.Close()
+
+	var schemaInfo []*mgmtv1alpha1.ConnectionConfigStatistics
+	for rows.Next() {
+		hasRows = true
```
nickzelei, line 103: I don't see where this is actually used anywhere?

nickzelei (changes requested):
left a few comments.

Ontop of this:
Please move these queries into sqlc and use the generated methods for them.

Does querying this table actually take into account the permissions of the user accessing the database?

backend/services/mgmt/v1alpha1/connection-service/connection.go
evisdrenova: i guess it doesn't, was just being careful and looking at a reference. it can just call it directly

backend/protos/mgmt/v1alpha1/connection.proto
```diff
@@ -53,12 +53,18 @@ message CheckConnectionConfigResponse {
...
-  repeated ConnectionConfigStatistics statistics = 3;
-}
-
-message ConnectionConfigStatistics {
-  string schema = 1;
-  int64 table_count = 2;
+  repeated ConnectionRolePrivileges privileges = 3;
+}
+
+message ConnectionRolePrivileges {
```
nickzelei, line 59: Minor thing but this should be singular as it represents a unary privilege. 
```suggestion
message ConnectionRolePrivilege {
```

backend/services/mgmt/v1alpha1/connection-service/connection.go
```diff
@@ -3,136 +3,151 @@ package v1alpha1_connectionservice
...
+  a. create user that has no access to a schema and validate - the connection should connect but no read/write to tables
+  b. check what happens when db schema doens't exist - for ex. maybe they want to create it later
+  c. create role that has some read/write privileges to some tables but not all
+  d. create role that has access to one schema but not another
+3. update docs for all of this
+4. update back end tests
+5. persist permissions in a separate tab (stretch)
+7. handle root user - since root user has access to everything, update to return checks if user is root
+*/
+
```
nickzelei, line 36: let's make sure this comment gets cleaned up prior to merge

frontend/apps/web/app/(mgmt)/[account]/connections/[id]/page.tsx
```diff
@@ -69,16 +69,17 @@ export default function ConnectionPage({ params }: PageProps) {
       </div>
     ),
   });
+
+  const basePath = `/${account?.name}/connections/${data?.connection?.id}`;
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/connections/[id]/page.tsx:73]: what's this?

backend/services/mgmt/v1alpha1/connection-service/connection.go
```diff
@@ -3,136 +3,151 @@ package v1alpha1_connectionservice
...
-	var hasRows bool // used to check if there are rows returned from the db
+		defer conn.Close()
 
-	switch req.Msg.ConnectionConfig.Config.(type) {
-	case *mgmtv1alpha1.ConnectionConfig_PgConfig:
-		rows, err = tx.QueryContext(cctx, pgQuery)
-	case *mgmtv1alpha1.ConnectionConfig_MysqlConfig:
-		rows, err = tx.QueryContext(cctx, mysqlQuery)
-	}
+		cctx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
nickzelei, line 65: This 5s timeout might be enough for ping, but it may not be enough for a larger db that takes a while to get all of the row results back since it may be a heavier query.

frontend/apps/web/app/(mgmt)/[account]/connections/[id]/page.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/connections/[id]/page.tsx:73]: i was starting to build out the permissions on a new tab and then decided to move it, left over code from that.

PR #1672, opened 2024-03-29
NEOS-947: add permissions page to postgres connections

evisdrenova:


frontend/apps/web/app/(mgmt)/[account]/connections/[id]/permissions/page.tsx
```diff
@@ -0,0 +1,211 @@
...
+  );
+
+  useEffect(() => {
+    if (
+      !pgConfig &&
+      data?.connection?.connectionConfig?.config.case === 'pgConfig'
+    ) {
+      setPgConfig(data.connection.connectionConfig.config.value);
+    }
+  }, [data, pgConfig]); //TODO: should listen to the fields in these objects but we don't know which ones will be filled out
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/connections/[id]/permissions/page.tsx:51]: Try to do this without the use effect as it's just going to cause issue with state refreshes.
See my comment down on line 59 for suggestion.

frontend/apps/web/app/(mgmt)/[account]/connections/[id]/permissions/page.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+  const {
+    data: validationRes,
+    isLoading: isLoadingValidation,
+    mutate: mutateValidation,
+  } = useTestProgressConnection(
+    account?.id ?? '',
+    pgConfig ?? new PostgresConnectionConfig({})
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/connections/[id]/permissions/page.tsx:59]: You can just auto calculate the pg config based on the incoming data without the need for a useEffect and a separate state variable (simpler is better here).

```suggestion
    data?.connection?.connectionConfig?.config.case === 'pgConfig' ? data.connection.connectionConfig.config.value : new PostgresConnectionConfig({})
```

This allows you to completely remove the useState and useEffect to have the same outcome.

nickzelei:
left a comment about simplifying but overall looks good!

PR #1688, opened 2024-04-01
Nick/neos 957 set more default values in helm charts

nickzelei:
Raises default limits on api, worker.
Disables autoscaling by default

### DIFF
--- a/worker/charts/worker/values.yaml
+++ b/worker/charts/worker/values.yaml
@@ -31,14 +31,14 @@ resources:
     cpu: "100m"
     memory: "128Mi"
   limits:
-    cpu: "300m"
-    memory: "384Mi"
+    cpu: "500m"
+    memory: "512Mi"
 
 autoscaling:
-  enabled: true
+  enabled: false
   minReplicas: 1
   maxReplicas: 10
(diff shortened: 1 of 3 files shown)


worker/charts/worker/values.yaml
```diff
@@ -31,14 +31,14 @@ resources:
...
     memory: "128Mi"
   limits:
-    cpu: "300m"
-    memory: "384Mi"
+    cpu: "500m"
+    memory: "512Mi"
 
 autoscaling:
-  enabled: true
+  enabled: false
```
nickzelei, line 38: users should have to opt in to autoscaling as this may be unexpected for basic deployments.

worker/charts/worker/values.yaml
```diff
@@ -31,14 +31,14 @@ resources:
...
   minReplicas: 1
   maxReplicas: 10
-  targetCPUUtilizationPercentage: 70
+  targetCPUUtilizationPercentage: 80
```
nickzelei, line 41: rolling this back to the k8s default

alishakawaguchi (approved):

PR #1843, opened 2024-04-28
NEOS-969: filter out defaults

evisdrenova:


backend/services/mgmt/v1alpha1/connection-data-service/connection-data.go
```diff
@@ -404,12 +406,20 @@ func (s *Service) GetConnectionSchema(
 
 		schemas := []*mgmtv1alpha1.DatabaseColumn{}
 		for _, col := range dbschema {
+			hasDefault := false
+			if col.ColumnDefault != nil {
+				// ColumnDefault is an interface we need to check the value to see if it's set or not
+				value := reflect.ValueOf(col.ColumnDefault)
+				hasDefault = !value.IsZero()
+			}
+
```
nickzelei, line 415: If you go into the `system.sql` file for postgres queries, there are two queries in there that can be updated that will make all of this code go away.

Look for this block (shows up twice):

```
    COALESCE(
        pg_catalog.pg_get_expr(d.adbin, d.adrelid),
        ''
    ) AS column_default,
```

That is the bit that calculates this column. However, sqlc isn't smart enough to know the type, so we just need to cast it to text like so:

```
    COALESCE(
        pg_catalog.pg_get_expr(d.adbin, d.adrelid),
        ''
    )::text AS column_default,
```

This will cause the Go struct to appear as a string. Then all of this reflect stuff can be removed.

backend/protos/mgmt/v1alpha1/connection_data.proto
```diff
@@ -66,6 +66,8 @@ message DatabaseColumn {
   string data_type = 4;
   // The isNullable Flag of the column
   string is_nullable = 5;
+  // Indicates where the column has a default value
+  bool has_default = 6;
```
nickzelei, line 70: I would suggest just returning the default value itself here instead of a computed. Since this is the database schema data, we might actually want to use that in the future. Then this can just be computed easily on the frontend.

```suggestion
  optional string column_default = 6;
```

frontend/apps/web/components/jobs/SchemaTable/transformer-handler.ts
```diff
@@ -74,10 +74,10 @@ function shouldIncludeSystem(
   filters: TransformerFilters
 ): boolean {
   if (
-    filters.hasDefault &&
+    !filters.hasDefault &&
     transformer.source === TransformerSource.GENERATE_DEFAULT
   ) {
-    return true;
+    return false;
```
nickzelei, line 80: I don't quite understand why this was changed.
We want this block to return true if the filter says that it has a default and the source is Generate default

nickzelei (changes requested):
thanks for working on this! Left a few comments.

backend/services/mgmt/v1alpha1/connection-data-service/connection-data.go
evisdrenova: Sweet - didn't know this was there. Updated! Also had to update a few other places that relied on that column being an interface

backend/protos/mgmt/v1alpha1/connection_data.proto
evisdrenova: yeah that can work - I was hesitant at first to do that because our front-end really doesn't touch much "actual" data right now, aka data that would be found in a column but the default value might be okay since I guess it's kinda like metadata? so in the future if we do a hybrid deployment, and we want to isolate any data to the backend/worker, then we would maybe have to reverse this to keep the front-end clean of any data. But maybe it's okay for now.

backend/services/mgmt/v1alpha1/connection-data-service/connection-data.go
```diff
@@ -370,11 +370,12 @@ func (s *Service) GetConnectionSchema(
...
-				Table:      col.TableName,
-				Column:     col.ColumnName,
-				DataType:   col.DataType,
-				IsNullable: col.IsNullable,
+				Schema:        col.TableSchema,
+				Table:         col.TableName,
+				Column:        col.ColumnName,
+				DataType:      col.DataType,
+				IsNullable:    col.IsNullable,
+				ColumnDefault: &col.ColumnDefault,
```
nickzelei, line 378: Think we want to do the same thing as below where we only assign ColumnDefault if the value is not empty, that way it is truely nil.

nickzelei (approved):
LGTM

PR #1894, opened 2024-05-02
docs: add troubleshooting guide and init schema guide

zackerydev:


docs/docs/guides/troubleshooting.md
```diff
@@ -0,0 +1,41 @@
...
+This is likely due to records in the source database, that are added _after_ the table on one side of the constraint is synchronized. If you have tables that are constantly in flux on the source database, and your job takes a long time to run this error happens more frequently.
+
+The easiest way to resolve it is to copy or snapshot your source connection to a fresh database with no other connections, and run the job again.
+
+### `ERROR: schema \"schema_name\" does not exist`
+
+Neosync does not support creating schemas on the fly. You must create the schema on the target database before running the job.
+
+
+### `unable to exec truncate cascade statements: ERROR: relation \"public.table_name\" does not exist`
```
nickzelei, line 31: This is interesting because if you unselect truncate as well as don't init, but your table doesn't exist, you'll get a different error in that it can't insert because the table doesn't exist :)

nickzelei (commented):
Amazing, thank you so much for adding this @zackerydev !

I'd like to merge this pretty much as is. I read thru and it all makes sense.

The only thing that is missing here is adding these two new pages to the `sidebars.ts` file so that they show up there. 
Once that is done, this can be merged in!

docs/docs/guides/troubleshooting.md
zackerydev: Yep - some of them are painfully obvious once you think about it

nickzelei:
Separately, I approved this PR for github action runs. Looks like spellcheck and prettier need to be fixed as well. Happy to help with that you run into any issues there.

docs/docs/guides/initializing-schemas.md
```diff
@@ -0,0 +1,47 @@
...
+
+This can potentially have some issues if you have a lot of migrations or if your migrations are not idempotent.
+
+If your migrations are any of the following you may run into issues:
+
+- Not Idempotent, meaning that running the migration multiple times (or against a completely blank/fresh database) will cause issues
+- Have hard coded database names (if you are choosing to run neosync against a separate logical database in your single instance)
+- Edit database users, global configurations or other settings that are not related to the schema/data
+
+Some of the above issues can be resolved by choosing to use a completely fresh database, or by running your migrations against a fresh database and then snapshotting that database to use as your target.
```
nickzelei, line 24: snapshotting seems like a valid word. 
You can exclude it for this file by adding an exclusion rule to the metadata.

```
# cSpell:words snapshotting
```

docs/docs/guides/troubleshooting.md
```diff
@@ -0,0 +1,41 @@
...
+
+This document is a collection of common issues you may face when running Neosync and how to resolve them.
+
+### `Sync: could not successfully complete sync activity: read tcp 0.0.0.0:42454->0.0.0.0:5432: i/o timeout`
+
+This is likely due to an overabundance of connections against the database. 
+
+You can resolve this by reducing the number of open connections in the "Connections" tab of Neosync. The default is at 80, the easiest way to debug is to start with something small like 10 and increase it until you hit the error again.
+
+### `could not successfully complete sync activity: pq: insert or update on table \"table_name\" violates foreign key constraint \"table_name_id_fkey\"`
```
nickzelei, line 20: Same here, this is valid given the context.

```
# cSpell:words fkey
```

zackerydev: I swapped this one to `foreign_key` seems likely that someone might add "fkey" accidentally else where in the file 👍🏻

docs/docs/guides/initializing-schemas.md
zackerydev: This sentence didn't make a ton of sense as I was reading it back so reworded it slightly

docs/docs/guides/troubleshooting.md
nickzelei: sounds good 👍

nickzelei (approved):
🚢

PR #1944, opened 2024-05-08
Nick/neos 1043 expose fk connection id as discrete field for rules based

nickzelei:


nickzelei:
also Closes NEOS-1004

PR #2021, opened 2024-05-24
Fixed links

KPCOFGS:
Fixed 2 links


KPCOFGS:
Hi! Thank you for your feedback! I see what you mean now. one reason for this might be caused by how GitHub and the website handled .md files differently. Is there a way to ensure the consistency between these two?

PR #2103, opened 2024-06-07
NEOS-1153: updated frontend and bff layer

evisdrenova:


backend/pkg/sqlconnect/sql-connector.go
```diff
@@ -427,7 +427,56 @@ func getGeneralDbConnectionConfigFromMysql(config *mgmtv1alpha1.ConnectionConfig
 			QueryParams: query,
 		}, nil
 	case *mgmtv1alpha1.MysqlConnectionConfig_Url:
-		return nil, nucleuserrors.NewNotImplemented("not currently implemented")
+		// follows the format [scheme://][user[:password]@]<host[:port]|socket>[/schema][?option=value&option=value...]
```
nickzelei, line 430: If you could add some tests in the sql-connector_Test file that thoroughly test this branch with various mysql connection strings, that would be great! :) There are some postgres equivalent tests that you can example from

backend/pkg/sqlmanager/mysql/mysql-manager.go
```diff
@@ -34,7 +34,8 @@ func (m *MysqlManager) GetDatabaseSchema(ctx context.Context) ([]*sqlmanager_sha
 	for _, row := range dbSchemas {
 		var generatedType *string
 		if row.Extra.Valid && row.Extra.String == "GENERATED" {
-			generatedType = &row.Extra.String
+			generatedTypeCopy := row.Extra.String
+			generatedType = &generatedTypeCopy
```
nickzelei, line 38: curious why this and the other examples needed to be changed here?

nickzelei (commented):
sweet! left a few comments.

backend/pkg/sqlmanager/mysql/mysql-manager.go
evisdrenova: linter was throwing an error due to an implicit memory aliasing - we were referencing a loop variable inside the for loop which can lead to all of the pointers pointing to the last item's data. 

````
pkg/sqlmanager/mysql/mysql-manager.go:37:20: G601: Implicit memory aliasing in for loop. (gosec)
                        generatedType = &row.Extra.String
```

Was doing a little [bit of reading](https://stackoverflow.com/questions/62446118/implicit-memory-aliasing-in-for-loop) and it mentioned that g0 1.22 solves this but golang-ci lint might be using an older version so it still throws the error. 

I guess we could just ignore it if we're confident that it's not a real warning. Happy to revert if that makes sense.

backend/pkg/sqlconnect/sql-connector.go
evisdrenova: sure thing

backend/pkg/sqlmanager/mysql/mysql-manager.go
nickzelei: all good, i'd trust the linter in this case! Might be fine, but could also be a hidden bug too.

Issue #2046, opened 2024-05-28
[NEOS-1126] Temporal Seed Script not working on some hosts


nickzelei:
There seems to be an issue with the Temporal seed script not working.

More details here: [https://discord.com/channels/1171894401140797591/1245086949547704401](https://discord.com/channels/1171894401140797591/1245086949547704401)

Either is an issue with Temporal or the CLI. May be due to host OS, it is unclear.

<sub>From [SyncLinear.com](https://synclinear.com) | [NEOS-1126](https://linear.app/nucleus-cloud/issue/NEOS-1126/temporal-seed-script-not-working)</sub>

PR #2131, opened 2024-06-13
NEOS-1126 fix(compose): pin admin-tools version to v1.23

nickrvieira:
This addresses the common issue of flag incompatibility between latest `temporalio/admin-tools` setup.

ISSUE REF: https://github.com/nucleuscloud/neosync/issues/2046

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -96,7 +96,7 @@ services:
 
   temporal-seed:
     container_name: neosync-temporal-seed
-    image: temporalio/admin-tools:latest
+    image: temporalio/admin-tools:1.23
     environment:
       - TEMPORAL_ADDRESS=temporal:7233
     entrypoint: /bin/sh -c "/create_schedules.sh"


compose.yml
```diff
@@ -96,7 +96,7 @@ services:
 
   temporal-seed:
     container_name: neosync-temporal-seed
-    image: temporalio/admin-tools:latest
+    image: temporalio/admin-tools:1.23
```
nickzelei, line 99: Hm, the temporal version we have set in the .env is `TEMPORAL_VERSION=1.22.6`
Should this be 1.22 then? Also open to bumping the temporal compose to the latest.

nickrvieira: 1.22 works as well - I just tried to find a middle ground but didn't realize that `.env` pinned this version

nickzelei: I see, ok. 
In the temporal compseo script I realized we actually have the admin tools defined, I commented them out a long time ago.

```
  # temporal-admin-tools:
  #   container_name: temporal-admin-tools
  #   depends_on:
  #     - temporal
  #   environment:
  #     - TEMPORAL_ADDRESS=temporal:7233
  #     - TEMPORAL_CLI_ADDRESS=temporal:7233
  #   image: temporalio/admin-tools:${TEMPORAL_VERSION}
  #   networks:
  #     - temporal-network
  #   stdin_open: true
  #   tty: true
```
Perhaps longterm I should go back to using that, and then interact with the container externally instead of doing what we are currently doing.
Either way, that is outside of the scope of this issue.

If the 1.23 image is working fine for you, let's go with it for now. Can figure out the better optimization later.

nickzelei: Good to know 1.24 has some issues. I will create a ticket and look into that.

nickzelei (approved):

compose.yml
nickzelei: FYI, created two followup issues: https://github.com/nucleuscloud/neosync/issues/2133 and https://github.com/nucleuscloud/neosync/issues/2134

nickzelei:
Thank you for the contribution!!

PR #2228, opened 2024-06-28
Experimenting with Tanstack Query

nickzelei:


backend/internal/cmds/mgmt/serve/connect/cmd.go
```diff
@@ -410,9 +411,16 @@ func serve(ctx context.Context) error {
 	}
 	mux.Handle("/", api)
 
+	c := cors.New(cors.Options{
```
nickzelei, line 414: TODO: Need to test this with same domains. May only need to allow this for development as in Staging, app and api are under the same root domain

Issue #2294, opened 2024-07-16
[NEOS-1237] Add github action that prevents merge on unlabeled PRs


nickzelei:
We should have a job that runs on a pull request that does not contain any labels (or maybe just always runs).

Then we can add a condition that it is unmergeable if it does not have a label attached.

We might need it to run all the time os that we can can "require" the test be passing before it is considered mergeable.

Could maybe be used as a reference: [https://stackoverflow.com/questions/71502652/prevent-merging-of-branch-conditionally-based-on-label](https://stackoverflow.com/questions/71502652/prevent-merging-of-branch-conditionally-based-on-label)

This way it forces us to label or PRs so that they are better associated in our changelog.

<sub>From [SyncLinear.com](https://synclinear.com) | [NEOS-1237](https://linear.app/nucleus-cloud/issue/NEOS-1237/add-github-action-that-prevents-merge-on-unlabeled-prs)</sub>

frankie-mur:
Heya @nickzelei , I could take this on 🫡

nickzelei:
> Heya @nickzelei , I could take this on 🫡

Sounds good!

PR #2306, opened 2024-07-17
workflow: add github action for checking pr label

frankie-mur:
This PR adds a new Github Action, the action checks if there is atleast one PR label present, if not it makes a comment in the PR. This PR addresses issue #2294
### Example: 
- Opening a PR without a label
image

- Adding a label after said comment
image

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -12,7 +12,7 @@ jobs:
   check_labels:
     runs-on: ubuntu-latest
     permissions:
-      pull-requests: write
+      issues: write
     env:
       HAS_LABELS: ${{ github.event.pull_request.labels[0] != null }}
     steps:
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -8,7 +8,8 @@ on:
       - opened
       - labeled
       - unlabeled
-
+permissions:
+  pull-requests: write
 jobs:
   check_labels:
     runs-on: ubuntu-latest
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,0 +1,35 @@
+name: Check PR Labels
+
+on:
+  pull_request:
+    branches:
+      - main
+    types:
+      - opened
+      - labeled
+      - unlabeled
+
+jobs:
+  check_labels:
+    runs-on: ubuntu-latest
+
+    env:
+      HAS_LABELS: ${{ github.event.pull_request.labels[0] != null }}
+
+    steps:
+      - name: Add a comment to the PR if no labels
+        if: env.HAS_LABELS == 'false'
+        uses: actions/github-script@v6
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'Your pull request does not have any labels. Please add at least one label to pass the checks.'
+            })
+
+      - name: Fail the job if no labels are found
+        if: env.HAS_LABELS == 'false'
+        run: exit 1
+        shell: bash
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -8,15 +8,13 @@ on:
       - opened
       - labeled
       - unlabeled
-permissions:
-  pull-requests: write
 jobs:
   check_labels:
     runs-on: ubuntu-latest
-
+    permissions:
+      pull-requests: write
     env:
       HAS_LABELS: ${{ github.event.pull_request.labels[0] != null }}
-
     steps:
       - name: Add a comment to the PR if no labels
         if: env.HAS_LABELS == 'false'
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,4 +1,4 @@
-name: Ensure PR Has Labels
+name: Ensure PR Has Label(s)
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


nickzelei (changes requested):
Hey @frankie-mur 
Overall, I think the PR looks great, thank you for jumping on this!

However, in the case where the job fails due to not having a label, it seems to be unable to add a comment due to a 403.
I'm assuming there is a way to detail in the config that the action runner should have permissions to write to the PR to comment?
If not, I think it's ok to simply fail the job instead of posting a comment.

```
RequestError [HttpError]: Resource not accessible by integration
    at /home/runner/work/_actions/actions/github-script/v6/dist/index.js:6842:21
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  status: 403,
  response: {
    url: 'https://api.github.com/repos/nucleuscloud/neosync/issues/2306/comments',
    status: 403,
    headers: {
      'access-control-allow-origin': '*',
      'access-control-expose-headers': 'ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset',
      'content-encoding': 'gzip',
      'content-security-policy': "default-src 'none'",
      'content-type': 'application/json; charset=utf-8',
      date: 'Wed, 17 Jul 2024 18:12:46 GMT',
      'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin',
      server: 'github.com',
      'strict-transport-security': 'max-age=31536000; includeSubdomains; preload',
      'transfer-encoding': 'chunked',
      vary: 'Accept-Encoding, Accept, X-Requested-With',
      'x-accepted-github-permissions': 'issues=write; pull_requests=write',
      'x-content-type-options': 'nosniff',
      'x-frame-options': 'deny',
      'x-github-api-version-selected': '2022-11-28',
      'x-github-media-type': 'github.v3; format=json',
      'x-github-request-id': '4441:9EC47:[19](https://github.com/nucleuscloud/neosync/actions/runs/9978397770/job/27578199328?pr=2306#step:2:20)42281:2FC5F4B:6698099E',
      'x-ratelimit-limit': '15000',
      'x-ratelimit-remaining': '14930',
      'x-ratelimit-reset': '1721242245',
      'x-ratelimit-resource': 'core',
      'x-ratelimit-used': '70',
Error: Unhandled error: HttpError: Resource not accessible by integration
      'x-xss-protection': '0'
    },
    data: {
      message: 'Resource not accessible by integration',
      documentation_url: 'https://docs.github.com/rest/issues/comments#create-an-issue-comment',
      status: '403'
    }
  },
  request: {
    method: 'POST',
    url: 'https://api.github.com/repos/nucleuscloud/neosync/issues/2306/comments',
    headers: {
      accept: 'application/vnd.github.v3+json',
      'user-agent': 'actions/github-script octokit-core.js/3.6.0 Node.js/[20](https://github.com/nucleuscloud/neosync/actions/runs/9978397770/job/27578199328?pr=2306#step:2:21).13.1 (linux; x64)',
      authorization: 'token [REDACTED]',
      'content-type': 'application/json; charset=utf-8'
    },
    body: '{"body":"Your pull request does not have any labels. Please add at least one label to pass the checks."}',
    request: { agent: [Agent], hook: [Function: bound bound register] }
  }
}
```

Also, is the job supposed to automatically re-run when a label is added? or does it only re-run on another commit? I added a label to this PR but didn't notice it re-running and had to retrigger it. Just asking.

nickzelei:
@frankie-mur 
I believe you need to configure the permissions block to allow the action to comment on the PR.
Can read more about it here: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs
We have these configured in a few different actions in this repo, but none for commenting. Could serve as a starting point though.

The action may have just not run because you are an outside collab. I have to manually approve. That could be it...because I was thinking the same thing you specified labeled and unlabeled. It might work differently if I open a PR myself.

.github/workflows/pull-request.yml
```diff
@@ -8,7 +8,8 @@ on:
       - opened
       - labeled
       - unlabeled
-
+permissions:
+  pull-requests: write
```
nickzelei, line 12: I think this belongs on the job, or rather, is where it would be better served!

```
jobs:
  check_labels:
    name: Check PR Labels
    runs-on: ubuntu-latest
    permissions:
        pull-requests: write
```

nickzelei:
@frankie-mur Looks like the job runs correctly when the label is added or removed! But still receiving a 403. I left a comment in the second where I believe the permissions block needs to be moved to!

frankie-mur:
@nickzelei appreciate the suggestion! I have moved the permission block over. Lets try that again 🤞

nickzelei:
ok @frankie-mur it still didn't work. I did a little research, and it appears we have the wrong permissions. 
What we actually need is `issues: write` as github treats PRs as a special type of issue and the ability to comment is actually underneath the issues section instead of pull requests (so confusing!!).

Anyways, maybe let's give that a go.

frankie-mur:
oh wow, that is a bit confusing. Thank you for you patience in this and also doing the (very) manual testing!  😂

nickzelei:
All good! I'm learning something too.
I'm not sure why, but it's still failing.

I wonder if it's really worth it at this point to have this. The job failing might be sufficient.

nickzelei:
I would assume the github token doesnt need to be explicitly specified? I read a little about the github client and it says it should find it from the context.

frankie-mur:
Thats disappointing! I have removed the add a comment job and made the job Name to be more specific to unblock this in the meanwhile...

nickzelei (approved):
LGTM! Thank you for this addition and for working through the changerequests!

PR #2323, opened 2024-07-20
NEOS-1252: Update Generate State transformer to support returning back a 2-letter state code or the full state name

evisdrenova:


worker/pkg/benthos/transformers/generate_state.go
```diff
@@ -10,11 +11,15 @@ import (
 // +neosyncTransformerBuilder:generate:generateState
 
 func init() {
-	spec := bloblang.NewPluginSpec().Description("Randomly selects a US state and returns the two-character state code.")
-
+	spec := bloblang.NewPluginSpec().Param(bloblang.NewBoolParam("state_code"))
```
alishakawaguchi, line 14: this should be optional and have a default set to false. also needs a top level description and param level description

backend/protos/mgmt/v1alpha1/transformer.proto
```diff
@@ -301,7 +301,9 @@ message GenerateSha256Hash {}
 
 message GenerateSSN {}
 
-message GenerateState {}
+message GenerateState {
+  bool state_code = 1;
```
nickzelei, line 305: This could use a better name besides just `state_code` as that isn't super clear what it means.

What does it mean if it's true, what does it mean if it's false? 
Please also add a comment here describing what this property does.

I'm assuming that the `false` value also retains the previous empty value?

worker/pkg/benthos/transformers/generate_state.go
```diff
@@ -10,11 +11,21 @@ import (
 // +neosyncTransformerBuilder:generate:generateState
 
 func init() {
-	spec := bloblang.NewPluginSpec().Description("Randomly selects a US state and returns the two-character state code.")
-
+	spec := bloblang.NewPluginSpec().Param(bloblang.NewBoolParam("state_code").Optional()).Description("Randomly selects a US state and either returns the two character state code or the full state name.")
```
nickzelei, line 14: Instead of using `Optional()` you can replace it with `Default(false)` to better describe the default behavior.

nickzelei: Also, if you check out some other generator like `generate_ssn`, you can see how the spec is formatted and how it includes a top level description as well as a description for each parameter. This provides well formed generated documentation for this generator.

backend/protos/mgmt/v1alpha1/transformer.proto
evisdrenova: im going to swap it actually for `generate_full_name` that way it maintains the current logic where if it's false (which is the default) it'll return the state code, otherwise users can opt into returning the full state name

PR #2352, opened 2024-07-25
NEOS-1254: update onboarding flow

evisdrenova:
I've removed the onboarding checklist in favor of a welcome modal. The welcome modal  will pop up for anyone who hasn't either completed it or skipped it. I was considering keeping the onboarding checklist but I didn't think it made much sense. If someone wants to skip the welcome modal, it's likely they don't want the onboarding checklist either. If someone completes the welcome modal, then they would go right into the job creation flow which would pretty much just do everything that the onboarding checklist wanted (with the exception of inviting other people but I don't htink thats a big deal). 

As part of this change, I've updated the AccountOnboardingConfig and removed a lot of the code that drove updating the onboarding checklist. For existing users, the pop-up will show since the function looks for the `hasCompletedOnboarding` key in the `onboarding_config` and that won't be present. This would mean that existing users who have already dismissed or completed the onboardingChecking will see the welcome modal pop-up and they can either skip it or just click through it. If we didn't want this then there are probably two options: 1. Keep the `onboarding_config` object as-is and just add on a new key, then update code to manage this where it checks for the presence of the current `onboarding_config` and then updates the `hasCompletedOnboarding` value, or 2. leave it as-is and just let people skip or complete the new modal. 

Thoughts @nickzelei 

Update:
- Added a loom video showing the flow. Clicking on the records button in the tab takes you to the usage page but running it locally and don't have usage so doesn't load but path is correct. 
https://www.loom.com/share/8ba9a12b4ad2441e8d8cdb98e6d5f3c1?sid=33d5b78b-9443-4733-ba9c-388ad3e48b9c


frontend/apps/web/components/onboarding-checklist/WelcomeDialog.tsx
```diff
@@ -0,0 +1,161 @@
...
+    getAccountOnboardingConfig,
+    { accountId: account?.id ?? '' },
+    { enabled: !!account?.id }
+  );
+  const queryclient = useQueryClient();
+  const { mutateAsync: setOnboardingConfigAsync } = useMutation(
+    setAccountOnboardingConfig
+  );
+
+  const [currentStep, setCurrentStep] = useState<number>(0);
```
nickzelei, line 49: Was thinking about suggesting that we move away form this numbered system and instead use a narrow string type and transition to different states. That would make this code a lot more readable. So instead of doing "currentStep + 1) it's instead `setStep('welcome')` `setStep('connect')` etc.

So instead of having `FormStep.name` be a wide string, instead make it something like this:

```
type FormStepName = "welcome" | "connect" | "configure" | "execute" | "router";
export interface FormStep {
  name: FormStepName;
  component: JSX.Element;
}
```

Then the currentStep becomes:
```
const [currentStep, setCurrentStep] = useState<FormStepName>('welcome');
```
A lot more readable.

frontend/apps/web/components/onboarding-checklist/WelcomeDialog.tsx
```diff
@@ -0,0 +1,161 @@
...
+          setCurrentStep={setCurrentStep}
+          setIsDialogOpen={setShowGuide}
+          completeForm={completeForm}
+        />
+      ),
+    },
+    {
+      name: 'connect',
+      component: (
+        <Connect currentStep={currentStep} setCurrentStep={setCurrentStep} />
```
nickzelei, line 68: Most of these forms don't need to know their step or current step. They just need to know `onNext()` or `onBack()`, `onSkip()`, etc, whichever is relevant. Makes this a lot more readable and easier to manage.

frontend/apps/web/components/onboarding-checklist/WelcomeDialog.tsx
```diff
@@ -0,0 +1,161 @@
...
+      });
+      queryclient.setQueryData(
+        createConnectQueryKey(getAccountOnboardingConfig, {
+          accountId: account.id,
+        }),
+        new GetAccountOnboardingConfigResponse({
+          config: resp.config,
+        })
+      );
+      setShowGuide(false);
```
nickzelei, line 116: just wanted to put in here that this kinda sucks if for whatever reason they can't hit our server to set the onboarding config to true. it will stay stuck on the screen until our server comes back online. Not much we can do about taht I suppose unless we were to cache it in local storage or something. Maybe that is a task for another day however.

backend/protos/mgmt/v1alpha1/user_account.proto
```diff
@@ -172,10 +172,7 @@ message SetAccountOnboardingConfigResponse {
 }
 
 message AccountOnboardingConfig {
-  bool has_created_source_connection = 1;
-  bool has_created_destination_connection = 2;
-  bool has_created_job = 3;
-  bool has_invited_members = 4;
```
nickzelei, line 178: This is a breaking change and properties in proto shouldnt ever be renamed or re-used.

So what you actually want to do here is to instead deprecate these by marking them as deprecated (you can see in other protos uses of the deprecated comment.

Then the `has_completed_onboarding` should be a new property that starts at `5`;

nickzelei: From the outside these can all just default to false and everything is gucci since the frontend will only care about the new property after this.

backend/sql/postgresql/models/models.go
```diff
@@ -1776,24 +1776,15 @@ func (r *RetryPolicy) FromDto(dto *mgmtv1alpha1.RetryPolicy) {
 }
 
 type AccountOnboardingConfig struct {
-	HasCreatedSourceConnection      bool `json:"hasCreatedSourceConnection"`
-	HasCreatedDestinationConnection bool `json:"hasCreatedDestinationConnection"`
-	HasCreatedJob                   bool `json:"hasCreatedJob"`
-	HasInvitedMembers               bool `json:"hasInvitedMembers"`
+	HasCompletedOnboarding bool `json:"hasCompletedOnboarding"`
```
nickzelei, line 1779: This is fine and shouldnt be a breaking change, however it may be worth thinking through the idea that if someone has already completed this v1 of onboarding, we it could be worth somehow marking this new one as true automatically.

We could also say fuck that are just require everyone see the new onboarding guide to keep it simple too.

nickzelei: Otherwise these DTO updates are just fine because the old properties will default to false if unspecified.

backend/protos/mgmt/v1alpha1/user_account.proto
nickzelei: You will also need to update the user service integration tests to accommodate your new changes.

nickzelei (changes requested):
OK there is a lot here, this is a big PR.

I didnt seen a loom demo so maybe you could attach one or just show me what this looks like tomorrow.

I left a number of comments. A little sketched out by some of the screenshots that I think are going to be present in the onboarding flow. It would be better I think if we could just use the components instead so they don't get out of date.

Either way I think this is enough for a first pass (might have comments on the welcome board itself..didnt really look at that at all with all of the other comments i left in this PR. Maybe we can just talk about that offline.

frontend/apps/web/components/site-header/SiteHeader.tsx
```diff
@@ -10,31 +11,42 @@ import {
...
     <header className="supports-backdrop-blur:bg-background/60 sticky top-0 z-50 w-full border-b dark:border-b-gray-700 bg-background dark:hover:text-white backdrop-blur">
       <div className="container flex h-14 items-center">
         <MainNav />
         <MobileNav />
         <div className="flex flex-1 justify-end items-center space-x-2">
-          <Upgrade
-            calendlyLink={systemAppConfig.calendlyUpgradeLink}
-            isNeosyncCloud={systemAppConfig.isNeosyncCloud}
-          />
+          {!systemAppConfig.isNeosyncCloud &&
```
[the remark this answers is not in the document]
evisdrenova: yeah just had this flipped so i could actually test it locally - will reset

frontend/apps/web/components/site-header/SiteHeader.tsx
```diff
@@ -10,31 +11,42 @@ import {
...
+            account?.type == UserAccountType.PERSONAL && (
+              <RecordsProgressBar identifier={account?.id ?? ''} />
+            )}
+          {systemAppConfig.isAuthEnabled && (
+            <Upgrade
+              calendlyLink={systemAppConfig.calendlyUpgradeLink}
+              isNeosyncCloud={systemAppConfig.isNeosyncCloud}
+            />
+          )}
+          {!systemAppConfig.isAuthEnabled && <AccountSwitcher />}
```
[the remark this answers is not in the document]
evisdrenova: same as above - was just for local testing to see the layout - will flip

frontend/apps/web/components/site-header/SiteHeader.tsx
```diff
@@ -10,31 +11,42 @@ import {
...
         <MainNav />
         <MobileNav />
         <div className="flex flex-1 justify-end items-center space-x-2">
-          <Upgrade
-            calendlyLink={systemAppConfig.calendlyUpgradeLink}
-            isNeosyncCloud={systemAppConfig.isNeosyncCloud}
-          />
+          {!systemAppConfig.isNeosyncCloud &&
+            account?.type == UserAccountType.PERSONAL && (
+              <RecordsProgressBar identifier={account?.id ?? ''} />
```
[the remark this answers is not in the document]
evisdrenova: updated with a new component that acts as a wrapper

frontend/apps/web/components/site-header/SiteHeader.tsx
```diff
@@ -10,31 +11,42 @@ import {
...
+            )}
+          {systemAppConfig.isAuthEnabled && (
```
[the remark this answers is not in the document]
evisdrenova: updated

frontend/apps/web/components/site-header/SiteHeader.tsx
```diff
@@ -1,3 +1,4 @@
+'use client';
```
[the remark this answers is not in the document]
evisdrenova: yeah agreed - reverted back to SSR and moved the records progress stuff to `<AccountsRecordProgress.tsx`

frontend/apps/web/components/onboarding-checklist/WelcomeOverview.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { ReactElement } from 'react';
+import { Button } from '../ui/button';
+import { ConnectDarkMode } from './ConnectDarkMode';
+import { ConnectLightMode } from './ConnectLightMode';
+
+interface Props {
+  currentStep: number;
+  setCurrentStep: (val: number) => void;
+  setIsDialogOpen: (val: boolean) => void;
+  completeForm: () => Promise<void>;
```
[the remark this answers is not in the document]
evisdrenova: yup updated to contrl from parent component using callbacks

frontend/apps/web/components/onboarding-checklist/WelcomeRouter.tsx
```diff
@@ -0,0 +1,175 @@
...
+import { ReactElement, useEffect, useState } from 'react';
+import { AiOutlineExperiment } from 'react-icons/ai';
+import { PageProps } from '../types';
+import { Button } from '../ui/button';
+
+interface Props {
+  currentStep: number;
+  setCurrentStep: (val: number) => void;
+  setIsDialogOpen: (val: boolean) => void;
+  completeForm: () => Promise<void>;
```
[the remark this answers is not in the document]
evisdrenova: updated

frontend/apps/web/components/onboarding-checklist/WelcomeDialog.tsx
evisdrenova: yeah this could a follow up where we store it in local storage and check that if the db isn't active

frontend/apps/web/components/site-header/RecordsProgressBar.tsx
```diff
@@ -12,13 +12,13 @@ import { dateToNeoDate, periodToDateRange, UsagePeriod } from '../usage/util';
 
 interface Props {
   identifier: string;
+  idtype: string;
```
nickzelei, line 15: Let's make this type safe.

here is an example of how we extract the case for the different connections.

```
// Helper function to extract the 'case' property from a config type
type ExtractCase<T> = T extends { case: infer U } ? U : never;

// Extraction type that pulls out all of the connection config cases
export type ConnectionConfigCase = NonNullable<
  ExtractCase<ConnectionConfig['config']>
>;
```

The same can be done for the metric identifiers.

nickzelei (commented):
overall looking good on the FE. Left a comment regarding the record count type.

backend/protos/mgmt/v1alpha1/user_account.proto
```diff
@@ -172,7 +172,16 @@ message SetAccountOnboardingConfigResponse {
...
-  bool has_completed_onboarding = 1;
+  // @deprecated - use has_completed_onboarding
+  bool has_created_source_connection = 1;
+  // @deprecated - use has_completed_onboarding
+  bool has_created_destination_connection = 2;
+  // @deprecated - use has_completed_onboarding
+  bool has_created_job = 3;
+  // @deprecated - use has_completed_onboarding
+  bool has_invited_members = 4;
+  // @deprecated - use has_completed_onboarding
```
nickzelei, line 183: I don't think you want this one to be deprecated 😉

nickzelei:
You'll need to merge in main or rebase and re-run the backend `make gen` to rebuild your proto changes with the latest.

Please note that the `*validate.go` files have been removes so do not re-add them.

nickzelei (approved):

Issue #2470, opened 2024-08-14
[NEOS-1323] sub text is different font size for synthetic data sample description


evisdrenova:
![image.png](https://uploads.linear.app/a9c4e27b-39a0-41a6-a85a-a5e6b97b33a2/d90709b7-901d-41bf-9109-a6cded38cf19/fd38a5a7-0c78-4250-991e-e1f9eddfbeef?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2E5YzRlMjdiLTM5YTAtNDFhNi1hODVhLWE1ZTZiOTdiMzNhMi9kOTA3MDliNy05MDFkLTQxYmYtOTEwOS1hNmNkZWQzOGNmMTkvZmQzOGE1YTctMGM3OC00MjUwLTk5MWUtZTFmOWVkZGZiZWVmIiwiaWF0IjoxNzIzNjAxNjAzLCJleHAiOjE3NTUxMzc2MDN9.ht1fq4aPTlvQHIZpfZJGlAOsiteH_gtiAQ_CigZ7fhI)

<sub>From [SyncLinear.com](https://synclinear.com) | [NEOS-1323](https://linear.app/nucleus-cloud/issue/NEOS-1323/sub-text-is-different-font-size-for-synthetic-data-sample-description)</sub>

frankie-mur:
Hey @evisdrenova , can I take a crack at this?

evisdrenova:
@frankie-mur go for it! Let me know if you have any questions.

PR #2488, opened 2024-08-16
feat: update SampleTable to use Form fields to match parent field

frankie-mur:
## Description 
- This PR updates the SampleTable.tsx file to fix text styling and sizing, this addresses the issue #2470. 
- I was not able to test the `sample` button press and calling openAPI with loading data into the table as I do not have an paid openAPI key and it seems free plan does not support :( , should not effect this but if someone can confirm that would be appreciated
- I am very novice with React/Tailwind so please feel free to give tips and pointers where needed 
- Thanks 👍


nickzelei:
Hi @frankie-mur ! 
Appreciate the PR.

Do you think it would be possible to update this component to use similar styles without using the Form components directly? This table isn't a part of the form itself so I would prefer if it did not use the form components.

frankie-mur:
Thanks for the review @nickzelei , I have updated much simpler and without using Form components

evisdrenova:
@frankie-mur here is the styling for the subtext for the other form fields, we'll just want them to look the same. 

```js
const FormDescription = React.forwardRef<
  HTMLParagraphElement,
  React.HTMLAttributes<HTMLParagraphElement>
>(({ className, ...props }, ref) => {
  const { formDescriptionId } = useFormField();

  return (
    <p
      ref={ref}
      id={formDescriptionId}
      className={cn('text-[0.8rem] text-muted-foreground', className)}
      {...props}
    />
  );
});
FormDescription.displayName = 'FormDescription';
```

frankie-mur:
Appreciate that @evisdrenova, should we also update the heading `Synthetic Data Sample` to match the other form fields?

nickzelei:
> Appreciate that @evisdrenova, should we also update the heading `Synthetic Data Sample` to match the other form fields? Also since we don't want use Form components directly would it be appropriate to use the same className attributes directly? ex.
> 
> ```
>      <p className="text-[0.8rem] text-muted-foreground tracking-tight">
>             A sample of synthetic data given the inputs above. Returns at most
>             10 records.
>      </p>
> ```

Yep it is totally appropriate to use the same class names! These are just tailwind classes so it's totally fine to use those as they are used for rendering purposes.

We could update the heading, however, it's not intended to stand out as it's a separate part of the form.
However, if it looks better uniform with the others, I'm down for it!

frankie-mur:
image

PR #2489, opened 2024-08-16
NEOS-1267 Add dynamodb support to Neosync CLI

alishakawaguchi:


backend/services/mgmt/v1alpha1/connection-data-service/connection-data.go
```diff
@@ -388,12 +389,101 @@ func (s *Service) GetConnectionDataStream(
...
+			}
+		}
+
 	default:
 		return nucleuserrors.NewNotImplemented(fmt.Sprintf("this connection config is not currently supported: %T", config))
 	}
 	return nil
 }
 
+func convertAttributeValueToDynamoDBJSON(av dynamotypes.AttributeValue) (map[string]any, error) {
```
nickzelei, line 430: Do we have this method somewhere else? Meaning: are we copying this anywhere or is this bespoke to the connection data? If so, we should centralize it...

cli/internal/benthos/inputs/neosync-connection-data.go
```diff
@@ -231,3 +253,93 @@ func (g *neosyncInput) Close(ctx context.Context) error {
...
+	result := make(map[string]any)
+	ktm := make(map[string]neosyncbenthos_dynamodb.KeyType)
+	for key, value := range item {
+		result[key] = convertDynamoDBValue(key, value, ktm)
+	}
+
+	return result, ktm
+}
+
+func convertDynamoDBValue(key string, value any, keyTypeMap map[string]neosyncbenthos_dynamodb.KeyType) any {
```
nickzelei, line 267: Is this specific to the CLI? Same question as the conn data.
Seems liek we should have the dyanmodb util stuff be centralized in a top level `interanl/dynamodb` module

nickzelei (approved):
LGTM, left two minor comments, but am cool if we push it off and just do a cleanup PR where we centralize a lot of the dynamo logic into one spot that we can test.

nickzelei:
Oh, one last thing.

The `func getCategory(cc *mgmtv1alpha1.ConnectionConfig) string {` function in `cli/internal/cmds/neosync/connections/list` should be updated to support DynamoDB connection types.

PR #2492, opened 2024-08-17
NEOS-1343: Fixes self-referencing logic in worker when following FK constraints

nickzelei:


nickzelei (commented):
Reviewed @alishakawaguchi 's changes, they look good.

alishakawaguchi (approved):
👍

PR #2581, opened 2024-09-01
NEOS-1379: handle duplicate mongo mappings

evisdrenova:


frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -696,8 +793,23 @@ function EditCollection(props: EditCollectionProps): ReactElement {
...
+  const conflictRows = data.filter(
+    (obj) =>
+      obj.column === currentColumn &&
+      `${obj.schema}.${obj.table}` !== currentSchemaTable
+  );
+
+  // filter available collections by excluding those that conflict
+  const availableRows = collections.filter(
+    (item) => !conflictRows.some((obj) => `${obj.schema}.${obj.table}` === item)
+  );
```
nickzelei, line 809: Since all of this is to just filter the collections, can we move this logic outside of this function to the caller and then just pass in the collections already filtered? That will keep this component simple and not require a leakage by passing in the entire data and index.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -639,24 +696,61 @@ function IndeterminateCheckbox({
...
+        `${item.schema}.${item.table}` === `${schema}.${table}`
+    );
+
+    const isDuplicate = selectedCollection.some(
+      (item: JobMappingFormValues) => item.column === val
+    );
+
+    if (!isDuplicate) {
+      setDuplicateError(false);
+      setInputValue(val);
```
nickzelei, line 732: looks like this is being called in both branches and should be moved outside of it for clarity.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -639,24 +696,61 @@ function IndeterminateCheckbox({
...
 
   const handleSave = () => {
     onEdit({ column: inputValue });
     setIsEditingMapping(false);
   };
 
+  const handleDocumentKeyChange = (
+    data: JobMappingFormValues[],
+    schema: string,
+    table: string,
```
nickzelei, line 718: These don't really need to be passed in to this function since they are provided via the parent. Removing these from this will simplify and not shadow the parent variables.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -639,24 +696,61 @@ function IndeterminateCheckbox({
 interface EditDocumentKeyProps {
   text: string;
   onEdit: (updatedObject: { column: string }) => void;
+  data: JobMappingFormValues[];
+  schema: string;
+  table: string;
```
nickzelei, line 701: I suggest actually removing the these props from this component and keeping the EditDocumentKey props much simpler by instead adding an `isDuplicate` callback function that this component can instead call. This will defer the logic to the parent.

I get squeamish whenever I see the `data` being passed around willy nilly because it's generally a heavy object that can be very large. We want to loop over it as minimally as possible. Keeping this component simply by instead having a simply `isDuplicate` function lets us further centralize the data usage.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -230,9 +262,23 @@ function AddNewRecord(props: AddNewRecordProps): ReactElement {
     context: {
       accountId: account?.id,
       isUserJavascriptCodeValid: validateUserJsCodeAsync,
+      data: data,
```
nickzelei, line 265: If we could not pass the data directly into the context as that is a pretty large object.
Let's keep the context simply by instead passing in a function that can be called by the context to check.

If you see my comments below about writing a single `isduplicate` function, that may be able to be either used here 1:1 or augments to handle both with shared logic.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -230,9 +262,23 @@ function AddNewRecord(props: AddNewRecordProps): ReactElement {
...
+
+  useEffect(() => {
+    const validateKeyField = async () => {
+      form.clearErrors('key'); // clear the errors on the key field - this is to make sure that the errors are cleared if the user switches collection values
+      await form.trigger('key'); // trigger validation
+    };
+    if (collectionValue !== '' && keyValue !== '') {
+      validateKeyField();
+    }
+  }, [collectionValue, keyValue]);
```
nickzelei, line 280: Instead of doing this, you should override the onChange handler on the collection input to simply reset the errors on the `key` whenever the collection changes.

That is more reacting and does not require this useEffect.

However, you'll probably want to take advantage of the `isDuplicate` function I've mentioned a few times throughout this PR to ensure that the collection that is being switched to does not have a duplicate `key` that might be in the form.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -190,19 +191,50 @@ interface AddNewRecordProps {
...
+
+        const selectedCollection: JobMappingFormValues[] =
+          context?.options?.context?.data.filter(
+            (item: JobMappingFormValues) =>
+              `${item.schema}.${item.table}` === collection
+          );
+
+        const isDuplicate = selectedCollection.some(
+          (item: JobMappingFormValues) => item.column === value
+        );
```
nickzelei, line 219: This can all go away and should be replaced by an `isDuplicate` function that I've mentioned throughout this PR.

nickzelei: Also, the `useForm` hook can take in a Context interface, which it looks like that was never created for the `AddNewNosqlRecordForm`. Probably worthwhile to create one. You can see a few other examples throughout the codebase for how to create this. Gives us better type safety.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -154,6 +154,7 @@ export default function NosqlTable(props: Props): ReactElement {
           <CardContent>
             <AddNewRecord
               collections={collections}
+              data={data}
```
nickzelei, line 157: I think once we rip through the other comments this shouldnt be needed anymore.

nickzelei (changes requested):
Left a few comments. Thanks for working this! I think we can clean this up to be a lot more efficient and require overall less code.
Let me know if you have any Q's.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -366,15 +486,27 @@ function AddNewRecord(props: AddNewRecordProps): ReactElement {
...
+  onDuplicate(row: Row): void;
   transformerHandler: TransformerHandler;
   onEdit(row: Row, index: number): void;
-  collections: string[];
-  onDuplicate(row: Row): void;
+  isDuplicateKey: (
+    newValue: string,
+    schema: string,
+    table: string,
+    currValue?: string
```
nickzelei, line 496: What is the purpose of the `currValue` here and why is it not enough to just ask: is the <schema, table, newKey> a duplication?

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -116,6 +122,50 @@ export default function NosqlTable(props: Props): ReactElement {
...
+
+  // useCallback ensures that we only re-run the function if the keySet changes
+  // we pass in an optional curr value so that we don't throw errors when searching for duplicates and the existing key is there
+  // for ex. if the initial value is "email" and we change it to "emaila" and then back to "email", it throws an error because the key set is made from all of the rows of the data which includes the original value, so we delete it if currValue is valid so we don't throw that error and allow the user to just save their current value without changing it
+  const isDuplicateKey = useCallback(
+    (newValue: string, schema: string, table: string, currValue?: string) => {
+      const key = `${schema}.${table}.${newValue}`;
+      if (currValue) {
+        keySet.delete(`${schema}.${table}.${currValue}`);
+      }
```
nickzelei, line 142: This is a little scary. something called `isDuplicateKey` I would not expect to modify the key set.
I would expect the `isDuplicateKey` to be really dumb and basically just be doing this:

```
return keySet.has(`${schema}.${table}.${newValue}`);
```
and nothing else.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -116,6 +122,50 @@ export default function NosqlTable(props: Props): ReactElement {
...
+      return keySet.has(key);
+    },
+    [keySet]
+  );
+
+  // used to calculate the collections that can be updated based on a given key value
+  // for ex. if a collection.key is "a.b.c" and we want to update it to "d.e.c" but "d.e.c" already exists, then we don't want to show the "d.e." collection as an uption for the update
+  const filteredCollections = useCallback(
```
nickzelei, line 150: a little confused as to what value this is providing.
I read your comment and I thought I understood it, but I couldnt repro what it was trying to accomplish.

I have something like this:

```
collection, key
---------------
col1, foo.bar
col2, foo.bar.baz
```
I was able to edit the first row and update `col1` to be `col2` - if that is what this is trying to prevent?

nickzelei (commented):
Left a few questions...some things I don't quite understand maybe you could help with.

Also, it would be a good sanity check if you could also make sure this stuff all works with Dynamodb.
Some of the string slicing is a little scary to me and would just be nice tomake sure it all is working for both mongo and dynamo.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -116,6 +122,50 @@ export default function NosqlTable(props: Props): ReactElement {
...
+      set.add(`${item.schema}.${item.table}.${item.column}`);
+    });
+    return set;
+  }, [data]);
+
+  // useCallback ensures that we only re-run the function if the keySet changes
+  // we pass in an optional curr value so that we don't throw errors when searching for duplicates and the existing key is there
+  // for ex. if the initial value is "email" and we change it to "emaila" and then back to "email", it throws an error because the key set is made from all of the rows of the data which includes the original value, so we delete it if currValue is valid so we don't throw that error and allow the user to just save their current value without changing it
+  const isDuplicateKey = useCallback(
+    (newValue: string, schema: string, table: string, currValue?: string) => {
```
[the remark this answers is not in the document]
evisdrenova: My thinking here is that if a user is trying to edit a mapping in line, then we have a `newValue` that we have to use as a key to check against the existing data to see if it's there. That value changes on every keystroke and it's not saved to the `data` object yet. 

So we use the newValue to create a string that we can then check to see if it exists in the set. If it does, then we return `true` for the duplicate. 

The currValue is optional here since it's only used for the inline mapping and is used to ensure that an existing mapping doesn't throw a value. 

If the data object is: `[{a.b.c},{d.e.f}]`, and the user clicks on the edit button, if we don't know what the currValue of the row that the user is trying to edit, then we throw an error if the user wants to not make any changes. For. ex. if the current row mapping is {a.b.c} and we want to edit this, the user makes a change to {a.b.t}, that works since that's not a mapping that exists. However, if the user, for whatever reason doesn't want to make the change and just wants to exist the mapping, then they will revert the mapping to {a.b.c} which will throw a duplicate error since that already exists. 

So the currValue is used to prevent this case by deleting the mapping that the user is currently editing from the set so that it doesn't throw a duplicate error and the user can save the original mapping. 

The other way we can fix this is to give the user a button to exit the editing process which will then end the editing process and not save the update. I didn't do that since it felt like too many buttons already in that small form.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
```diff
@@ -253,7 +349,26 @@ function AddNewRecord(props: AddNewRecordProps): ReactElement {
...
+                      const lastDotIndex = value.lastIndexOf('.');
+                      const schema = value.substring(0, lastDotIndex);
+                      const table = value.substring(lastDotIndex + 1);
+                      if (isDuplicateKey(currentKey, schema, table)) {
+                        form.setError('key', {
+                          type: 'manual',
+                          message:
+                            'This key already exists in the selected collection.',
+                        });
+                      }
```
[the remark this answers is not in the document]
evisdrenova: This is used to catch the scenario where the user selects the collection in the new mapping form`a` and then enters  key 'b'. It returns back a `true` for 'isDuplicateKey()'. Then they try to change the collection to `b`. If collection `b` already has key 'a', then we need to throw the `isDuplicateKey()` error here. But since the uniqueMapping sits on the `key` field, it doesn't do that since the key field isn't changing, it's the collection field. 

I suppose we could move this into a `.(test)` function on the collection. I haven't tried it and my initial thought was to do it on the select, so it might work. Would need to test it.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
evisdrenova: answered above

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
evisdrenova: no it's catching that if the collection.key already exists in the data, then we don't want to present that collection to the user as an option that they can switch to.

for ex. say you have this:

collection, key
---------------
col1, foo.bar
col2, foo.bar

and you want to change `col2`, we want to filter out `col1` out of the list of collections that they could change `col2` to so they don't create a duplicate mapping.

But that is another thing we need to consider, and this PR doesn't answer that. This PR prevents duplicate mappings at the leaf level. Do we want to prevent duplicate mappings at any point in the path level? i.e. what you mentioned, since that woudl overwrite the value at `foo.bar.baz` with `foo.bar`

nickzelei: ah okay, I think it was this comment `then we don't want to show the "d.e." collection as an uption for the update` that threw me off. 

makes sense.

I think having mappings for `foo.bar` and `foo.bar.baz` is totally fine though.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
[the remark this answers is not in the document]
nickzelei: I see your thinking.

However, this is really just an edge case when the `newValue` and `currValue` are equivalent for the existing collection right?

onDuplicate should be smart enough to know that for the current rowIndex being updated, if the values are the same, don't do anything.

onDuplicate could just not fire if it knows `curValue === newValue`.

Anyways, just a thought, maybe I'm missing something here, but my spidey senses tend to go off when I see mutations in functions that are just checking if something exists in a set.

nickzelei (approved):
Overall the functionality is there but still have some thoughts around the handling of the onDuplicate and isDuplicate functions.

frontend/apps/web/components/jobs/NosqlTable/NosqlTable.tsx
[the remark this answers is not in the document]
evisdrenova: will handle this in a new PR and clean it up a little more

PR #2588, opened 2024-09-03
NEOS-1338: Add support for checking the connection in the connect page

evisdrenova:
This adds in real time connection validation into the Connect page as the user selects a connection. It aims to solve/avoid two problems (that we've come across):

1. People not testing their connections and then getting to the schema page and not understanding why their schema isn't loading. 
2. Something going wrong with a connection at some point in time before they configure the job and then they come back to it and the connection is no longer valid. 

Similar to the test connection modal, possible responses are:

1. Success
2. Warning -> orange stop sign, usually means that we were able to connect but don't see anything. Also includes a link to the permissions table for more info. 
3. Errors -> unable to connect, takes them to the permissions table which has a longer explanation of the issue

Warnings and Errors don't stop the user from moving forward, just tell the user that something is potentially wrong. 

Demo:
https://www.loom.com/share/36d5b9edf19745bdbcf4c7312ba5c430?sid=6b80f535-4046-482f-8572-9f8dc99af3b3


backend/protos/mgmt/v1alpha1/connection.proto
```diff
@@ -48,6 +48,19 @@ message CheckConnectionConfigRequest {
   ConnectionConfig connection_config = 1;
 }
 
+message CheckConnectionConfigByIdRequest {
+  // The connection ID whose connection we want to validate
```
nickzelei, line 52: Maybe something simpler like this? the `whose connection` is a little strange since a connection isn't really a person.
```suggestion
  // The connection id to validate
```

backend/services/mgmt/v1alpha1/connection-service/connection.go
```diff
@@ -163,6 +163,51 @@ func (s *Service) CheckConnectionConfig(
...
+
+	_, err = s.verifyUserInAccount(ctx, nucleusdb.UUIDString(conn.AccountID))
+	if err != nil {
+		return nil, err
+	}
+
+	dtoVal, err := dtomaps.ToConnectionDto(&conn)
+	if err != nil {
+		return nil, err
+	}
```
nickzelei, line 190: All of this logic can be removed and replaced with an internal call to `GetConnection`. 

Then this function just becomes:

```
connResp, err := s.GetConnection(ctx, ...)
if err != nil {
  return nil, err
}

... your other check code below ...
```

backend/services/mgmt/v1alpha1/connection-service/connection.go
```diff
@@ -163,6 +163,51 @@ func (s *Service) CheckConnectionConfig(
...
+
+	resp, err := s.CheckConnectionConfig(ctx, checkReq)
+	if err != nil {
+		return nil, err
+	}
+
+	return connect.NewResponse(&mgmtv1alpha1.CheckConnectionConfigByIdResponse{
+		IsConnected:     resp.Msg.IsConnected,
+		ConnectionError: resp.Msg.ConnectionError,
+		Privileges:      resp.Msg.Privileges,
```
nickzelei, line 206: It's good practice to use the DTO Getters instead of accessing the properties directly, at least when the props are non-nil.
```suggestion
		IsConnected:     resp.Msg.GetIsConnected(),
		ConnectionError: resp.Msg.GetConnectionError(), // maybe don't need this one if it is a pointer and go is unhappy with the getter
		Privileges:      resp.Msg.GetPrivileges(),
```

backend/protos/mgmt/v1alpha1/connection.proto
```diff
@@ -48,6 +48,19 @@ message CheckConnectionConfigRequest {
   ConnectionConfig connection_config = 1;
 }
 
+message CheckConnectionConfigByIdRequest {
```
nickzelei, line 51: Thank you for adding comments!

backend/backend/internal/mocks/github.com/prometheus/client_golang/api/prometheus/v1/mock_API.go
```diff
@@ -0,0 +1,1376 @@
+// Code generated by mockery. DO NOT EDIT.
```
nickzelei, line 1: Hm, why is this in here?

frontend/apps/web/app/(mgmt)/[account]/new/job/connect/page.tsx
```diff
@@ -576,3 +718,76 @@ export default function Page({ searchParams }: PageProps): ReactElement {
...
+  id: string | undefined;
+  accountName: string;
+}
+
+function TestConnectionBadge(props: TestConnectionBadgeProps): ReactElement {
+  const { isValidating, validationResponse, id, accountName } = props;
+
+  const url = `/${accountName}/connections/${id}/permissions`;
+
+  const handleValidationResponse = (
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/new/job/connect/page.tsx:734]: Probably worth moving this outside of this function to be its own standalone component that is isolated, that way it doesnt have to be rebuilt every time and react can better render it.

frontend/apps/web/app/(mgmt)/[account]/new/job/connect/page.tsx
```diff
@@ -576,3 +718,76 @@ export default function Page({ searchParams }: PageProps): ReactElement {
...
+    }
+  };
+
+  return (
+    <div className="relative flex items-center">
+      <div className="absolute">
+        {id && isValidating ? (
+          <Spinner className="text-black dark:text-white" />
+        ) : id && !isValidating ? (
+          <div>{handleValidationResponse(validationResponse)}</div>
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/new/job/connect/page.tsx:786]: let's move that internal component to be its own module-level function.
```suggestion
          <div><TestConnectionValidationResponse ...props... />
```

frontend/apps/web/app/(mgmt)/[account]/new/job/connect/page.tsx
```diff
@@ -144,156 +178,194 @@ export default function Page({ searchParams }: PageProps): ReactElement {
...
+                                  mongodb={mongodb}
+                                  dynamodb={dynamodb}
+                                  mssql={mssql}
+                                  newConnectionValue={NEW_CONNECTION_VALUE}
+                                />
+                              </SelectContent>
+                            </Select>
+                          </div>
+                          <div>
+                            <TestConnectionBadge
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/new/job/connect/page.tsx:355]: I watched the loom video. I think this component itself looks fine, but the error and warnings states are really long and kind of stick out.
image
Is there any way we could make this smaller or have it appear on the line below?
Or perhaps just be an icon that has a hover tooltip?
What does this look like on smaller screen sizes?

nickzelei (changes requested):
Nice! This will undoubtedly be useful. Left a few comments for cleaning up as well as a few questions.

backend/protos/mgmt/v1alpha1/connection.proto
evisdrenova: much better :)

backend/backend/internal/mocks/github.com/prometheus/client_golang/api/prometheus/v1/mock_API.go
evisdrenova: hm no idea? i ran `mockery` from the cli when i was trying to create a mock for testing but not sure why that got picked up? Shall i delete it?

nickzelei: Hm, maybe try deleting this.
I just ran mockery from the tip of main and saw no changes, so it's unclear to me why this is here.

frontend/apps/web/components/ui/form.tsx
```diff
@@ -153,14 +154,17 @@ const FormMessage = React.forwardRef<
...
+      <MdErrorOutline />
+      <p
+        ref={ref}
+        id={formMessageId}
+        className={cn('text-xs font-medium ', className)}
+        {...props}
+      >
+        {body}
+      </p>
+    </div>
```
nickzelei, line 167: Hey this is a shadcn component and generally shouldn't be touching stuff in here in an effort to keep it pretty upstream with shadcn so that it makes it easy to pull  in future updates.

In other words, why the change to this now? And does this styling look okay throughout the rest of the app for all of our forms that use FormMessage ?

frontend/apps/web/app/(mgmt)/[account]/new/job/connect/page.tsx
```diff
@@ -341,172 +412,252 @@ export default function Page({ searchParams }: PageProps): ReactElement {
...
+                                            `destinations.${index}.connectionId`
+                                          )
+                                        ]?.isValidating && (
+                                          <Spinner className="text-black dark:text-white absolute" />
+                                        )}
+                                    </div>
+                                  </div>
                                 )}
                               </FormControl>
+                              <div>
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/new/job/connect/page.tsx:628]: I think if you add a `flex` class, this might look a little bit nicer as it will only take up the space it needs.
image

Also, it looks like the corners are rounded on the success state but not on the other state. Maybe that could be updated for consistency.

Comment (nickzelei) [reply frontend/apps/web/app/(mgmt)/[account]/new/job/connect/page.tsx:628]: One thing we might want to consider too is that for ephemeral environments this will always be red. But maybe that's ok.

nickzelei (commented):
left a few comments, overall it looks really good!

One thing for us to also consider is that there currently isn't a way for this to be reactive.

Meaning: it would be great if I saw that the connection was failing...fixed it...and then when I came back to the page it re-triggered (if this was a hook) or I had a way to manually re-check the connection via a button click.
Right now it only does it on change.

For source it is easy...because it's only one so that could easily be a hook. Gets a little bit more hairy with the multiple destinations.

On a separate note: Be aware that both the AIGenerate and Generate have their own connection pages that this isn't currently being applied to. Maybe that could be done as a follow up?

frontend/apps/web/components/ui/form.tsx
evisdrenova: yeah i updated it since it looked different compared to the other warning that we show in the connect page now. I checked in other places and I (personally) think it looks much better.

evisdrenova: yeah i updated it since it looked different compared to the other warning that we show in the connect page now. I checked in other places and I (personally) think it looks much better. But good catch, I only want it to be applied to error cases so will add a check here for that.

frontend/apps/web/app/(mgmt)/[account]/new/job/connect/page.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/new/job/connect/page.tsx:628]: yeah i think that's okay, also why i didn't make it invalidate the form and the ability to proceed. but yes the flex class sounds good!

Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/new/job/connect/page.tsx:628]: updated :) wish tailwind could pick up `rrounded-xl` or i just need to learn how to spell...

evisdrenova:
yeah i considered that but didn't want to shove it all into one PR. I also need to add this to the source and destination pages in the job/[id] as well + a few other fast follows

nickzelei (approved):
🚢

Issue #2591, opened 2024-09-03
[NEOS-1396] Add support for double quotes in schema and tables for Postgres


nickzelei:
We have support for special characters, however, this breaks down when building many of our queries because double quotes need to be escaped properly, which is done by adding an extra double quote.

```
CREATE TABLE "user""s_data@" (id TEXT NOT NULL PRIMARY key);
```

This creates a table called `user"s_data@`

Goqu only knows how to wrap tables in the correct delimiter (for Postgres, it is a double quote), but in order for the statement to be valid, we have to add in an extra `"` wherever we find one.

This change will have to be made anywhere that we are using Goqu.

Well known places:

* sqlmanager_postgres
* querybuilder2

Keep an eye out for:

[Add support for escaping double quotes in Postgres Identifiers · Issue #428 · doug-martin/goqu](https://github.com/doug-martin/goqu/issues/428)

<sub>From [SyncLinear.com](https://synclinear.com) | [NEOS-1396](https://linear.app/nucleus-cloud/issue/NEOS-1396/add-support-for-double-quotes-in-schema-and-tables-for-postgres)</sub>

PR #2592, opened 2024-09-03
NEOS-1393: adds support for postgres special characters except for double quotes

nickzelei:
Adds support for special characters in postgres schemas and tables.

This does not currently support schemas/tables with double quotes.

Fast follow in Github: https://github.com/nucleuscloud/neosync/issues/2591

### DIFF
--- a/backend/pkg/dbschemas/sql/postgresql/queries/system.sql
+++ b/backend/pkg/dbschemas/sql/postgresql/queries/system.sql
@@ -256,12 +256,12 @@ ORDER BY
 SELECT
     con.conname AS constraint_name,
     con.contype::TEXT AS constraint_type,
-    con.connamespace::regnamespace::TEXT AS schema_name,
+    nsp.nspname AS schema_name,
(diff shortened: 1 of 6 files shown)


backend/pkg/dbschemas/sql/postgresql/queries/system.sql
```diff
@@ -256,12 +256,12 @@ ORDER BY
 SELECT
     con.conname AS constraint_name,
     con.contype::TEXT AS constraint_type,
-    con.connamespace::regnamespace::TEXT AS schema_name,
+    nsp.nspname AS schema_name,
```
nickzelei, line 259: Using the table from nsp.nspname as the `con.connamespace::regnamespace::TEXT` added an extra quote that is only needed in some cases and we mostly want to work with unescaped values.

alishakawaguchi (approved):

nickzelei:
Integration test that failed must be flaky. Ran a few times locally and no failures.

PR #2593, opened 2024-09-03
updates duplicate mapping functinality to be more performant

evisdrenova:
A few updates to make making mongo mappings more performant. 

Abstracts the onDuplicate() function into a callback that gets called in the parent component.

Reworks the unique mapping creation to create a unique identifier instead of using an int.


nickzelei (approved):
🚢

PR #2597, opened 2024-09-03
NEOS-1384 Fix mssql default

alishakawaguchi:


worker/pkg/workflows/datasync/activities/gen-benthos-configs/sync.go
```diff
@@ -507,6 +503,42 @@ func (b *benthosBuilder) getSqlSyncBenthosOutput(
...
+	identityColumns []string,
+	colTransformerMap map[string]map[string]*mgmtv1alpha1.JobMappingTransformer,
+) (prefix, suffix *string) {
+	var pre, suff *string
+	if driver != sqlmanager_shared.MssqlDriver || len(identityColumns) == 0 {
+		return pre, suff
+	}
+	tableName := neosync_benthos.BuildBenthosTable(schema, table)
+	if hasPassthroughIdentityColumn(tableName, identityColumns, colTransformerMap) {
+		p := fmt.Sprintf("SET IDENTITY_INSERT %s ON;", tableName)
```
nickzelei, line 517: Should these table names be escaped?

worker/pkg/benthos/sql/output_sql_insert.go
```diff
@@ -255,28 +268,86 @@ func (s *pooledInsertOutput) WriteBatch(ctx context.Context, batch service.Messa
...
+// use when all columns are identity generation columns
+func getMssqlDefaultValuesInsertSql(schema, table string, rowCount int) string {
+	var sql string
+	for i := 0; i < rowCount; i++ {
+		sql += fmt.Sprintf("INSERT INTO %s.%s DEFAULT VALUES;", schema, table)
+	}
+	return sql
+}
+
+func filterIdentityColumns(
```
nickzelei, line 310: nit: Is this thing returning cols without identity-specific columns? Kinda what it looks like from the code but not immediately clear from the func name.

Also not obvious that it only seems to do so if the arg is DEFAULT. But maybe it is always that? 
Maybe a func comment could be helpful here.

nickzelei (approved):
overall lgtm ,left a few comments about escaping schema and tablenames

PR #2599, opened 2024-09-04
Fix Mssql subset not saving

alishakawaguchi:


nickzelei (approved):

nickzelei (commented):
Something integration tests will catch eventually ;)

PR #2600, opened 2024-09-04
NEOS-1400: Enable sql server truncate

alishakawaguchi:


nickzelei (approved):
one missing check on the missing destdb close, but otherwise lgtm! nice!!

PR #2602, opened 2024-09-04
Feature extra env init container

kenzaboussaoud8:
This PR adds support for the extraEnvVars section to the migrations section of the api.

### DIFF
--- a/backend/charts/api/templates/deployment.yaml
+++ b/backend/charts/api/templates/deployment.yaml
@@ -80,7 +80,7 @@ spec:
             - migrate
             - up
 
-          {{- with .Values.extraEnvVars }}
+          {{- with .Values.migrations.extraEnvVars }}
           {{- range .}}
           - name: {{ .name }}
             {{- if .value }}
--- a/backend/charts/api/templates/deployment.yaml
+++ b/backend/charts/api/templates/deployment.yaml
@@ -80,6 +80,31 @@ spec:
             - migrate
             - up
 
+          {{- with .Values.extraEnvVars }}
+          {{- range .}}
+          - name: {{ .name }}
+            {{- if .value }}
+            value: {{ .value | quote }}
+            {{- else if .valueFrom }}
+            valueFrom:
+              {{- if .valueFrom.secretKeyRef }}
+              secretKeyRef:
+                name: {{ .valueFrom.secretKeyRef.name }}
+                key: {{ .valueFrom.secretKeyRef.key }}
+              {{- end }}
+              {{- if .valueFrom.configMapKeyRef }}
+              configMapKeyRef:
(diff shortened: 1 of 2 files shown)


backend/charts/api/templates/deployment.yaml
```diff
@@ -80,6 +80,31 @@ spec:
             - migrate
             - up
 
+          {{- with .Values.migrations.extraEnvVars }}
```
nickzelei, line 83: Hey so this isn't quite right as-is. 
The init containers doesn't have any env vars defined on it by default here so in order for this to generate valid yaml you'll have to add in the `env:` keyword that this will be nested under.

however, not sure if k8s likes having a potentially empty `env:` defined.
We could add an env var that is always there similar to theother container like this:

```
          env:
            - name: HOST_IP
              valueFrom:
                fieldRef:
                  fieldPath: status.hostIP
```
That way the `extraEnvVars` can just be conditionally added underneath of that, which makes the logic a lot simpler.

nickzelei (changes requested):
Hey, thanks for the PR! Left a comment.

You'll also want to run `make helm/docs` to generate the helm documentation and get that added to the commit.

nickzelei:
Separately, curious as to what your usecase is for this feature addition?

kenzaboussaoud8:
> Separately, curious as to what your usecase is for this feature addition?

Hello Nick, this is to simplify our deployment process by reducing the need for manual patches. Specifically, we need to avoid applying patches for db credentials.

nickzelei:
> > Separately, curious as to what your usecase is for this feature addition?
> 
> Hello Nick, this is to simplify our deployment process by reducing the need for manual patches. Specifically, we need to avoid applying patches for db credentials.

Hmm, still not super clear. The db migrations container is fully configurable via the helm chart through the values. What are you looking to specify in the extra that isn't already available to configure through the helm chart as-is today?

backend/charts/api/templates/deployment.yaml
```diff
@@ -79,6 +79,36 @@ spec:
           args:
             - migrate
             - up
+          env:
+            - name: HOST_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+
+          {{- with .Values.migrations.extraEnvVars }}
```
nickzelei, line 88: I tried running this but the helm output failed due to indentation I believe. I think it might need to be indented one more section so that it appears underneath the `env`.

nickzelei:
Hey @kenzaboussaoud8 - I went ahead and took this over the line. Since outside collabs branches are slightly different, i wasn't able to push to this pull request directly.
You can see my final PR here: https://github.com/nucleuscloud/neosync/pull/2717

Thank you for the contribution!

PR #2613, opened 2024-09-05
Updates Docs to latest Docusaurus version

nickzelei:


docs/src/theme/Blog/Components/Author/index.tsx
```diff
@@ -0,0 +1,93 @@
+import Link, { type Props as LinkProps } from '@docusaurus/Link';
```
nickzelei, line 1: Had to add this to edit the styles.module.css as it is currently broken in the latest version...

nickzelei: Otherwise this component is unmodified.

alishakawaguchi (approved):

PR #2632, opened 2024-09-06
NEOS-1415: re enable team accounts in oss

nickzelei:


alishakawaguchi (approved):

backend/services/mgmt/v1alpha1/user-account-service/users.go
```diff
@@ -230,6 +230,9 @@ func (s *Service) CreateTeamAccount(
 	if !s.cfg.IsAuthEnabled {
 		return nil, nucleuserrors.NewForbidden("unable to create team account as authentication is not enabled")
 	}
+	if s.cfg.IsNeosyncCloud {
+		return nil, nucleuserrors.NewForbidden("creating team accounts in Neosync Cloud is not currently enabled. Please contact us to gain access to a team account.")
```
evisdrenova, line 234: The first makes it sound like we don't support it at all when in reality they just need to contact us for it. 

Maybe just say, "Please contact us to create a team account." 

Eventually, it'll just allow them to upgrade directly to it once the upgrade flow is set up.

nickzelei: Sure, this was moreso to avoid someone circumventing the UI

PR #2637, opened 2024-09-08
NEOS-1403:update forms to not shift on errors

evisdrenova:
This updates styling across most of the forms in the app to reduce the "jumping" of the UI when there is an error. It gives the <FormMessage /> component a static height, even if there is no error, so when there is an error it just fills the space without causing the rest of the UI to shift down.

### DIFF
--- a/frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx
+++ b/frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx
@@ -226,12 +226,12 @@ export default function NewTransformer(): ReactElement {
                     </SelectContent>
                   </Select>
                 </FormControl>
-                {/* <FormMessage />  we don't have any validation on this field so there's nothing to render*/}
+                <FormMessage />
               </FormItem>
             )}
           />
           {formSource != null && formSource !== 0 && (
-            <div className="pt-8">
+            <div>
               <FormField
                 control={form.control}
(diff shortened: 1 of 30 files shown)


frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx
```diff
@@ -226,12 +226,12 @@ export default function NewTransformer(): ReactElement {
                     </SelectContent>
                   </Select>
                 </FormControl>
-                <FormMessage />
+                {/* <FormMessage />  we don't have any validation on this field so there's nothing to render*/}
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx:229]: Nit: Hm, does this change the rendering? If so, then it kinda seems like `<Formmessage />` should be updated to handle that better. 
That way if this ever gets validation in the future it isn't missing its message.

nickzelei (approved):
Hm, lots of style changes, looking forward to seeing what it looks like.
Would be great for style-focused PRs to have a short gif or something to see it in action.

frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx]: it does ever so slightly. I added it back in and then just updated the padding on the component underneath it so it's not awkwardly padded.

PR #2638, opened 2024-09-08
NEOS-1423: make transformer slide out controlled

evisdrenova:


frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx
```diff
@@ -108,7 +119,8 @@ export default function EditTransformerOptions(props: Props): ReactElement {
                   ...value,
                   config: convertTransformerConfigToForm(newval),
                 });
-                setIsSheetOpen(false);
+                // setIsSheetOpen(false);
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx:122]: remove commented code.

frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx
```diff
@@ -55,15 +54,24 @@ interface Props {
...
+
+  const handleClose = () => {
+    setIsSheetOpen(false);
+  };
+
   return (
-    <Sheet open={isSheetOpen} onOpenChange={setIsSheetOpen}>
+    <Sheet
+      open={isSheetOpen}
+      onOpenChange={(open) => open && setIsSheetOpen(open)}
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx:65]: The way I'm reading this is that `onOpenChange` is changed to only work in a single direction now? So this will only fire setIsSheetOpen if `open` is true.

This seems unorthodox given the way the sheet is intended to be defined, so could you leave a short code comment explaining why it is being done this way? It seems like the onClick on :74 is handling the initial open, which would explain it.

This `EditTransformerOptions` is used in a few different spots and if I remember correctly, it only has troubles with auto-closing in the job mappings table. The bulk transformer sheet doesn't have the issue this PR is fixing.
image

Were these changes tested in the places this thing is used to ensure it works as intended?

nickzelei (approved):
Left a few minor comments, overall I think this looks good, however my main concern is that the sheet itself is actually totally fine as it works as intended everywhere except the job mappings table (which is a bit of a larger refactor). The main problem I think is that when you click off your screen and re-focus, the sheet closes unintentionally.

No doubt though having the sheet not close when clicking off of it is an improvement (via the onPointerDownOutside).

So overall, if this PR solves that issue without affecting the other uses, LGTM.

frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx]: Ah good catch yeah 74 sets it to be open so not need to set it on the onOpenChange, updated. 

The bulk transformer in the job mappings has the same problem. If you click outside of the form, it will auto-close. With this change, the bulk transformer, nosql and job mappings are all behaving like I want them to (not closing when you click outside of the form).

PR #2640, opened 2024-09-09
NEOS-1421, NEOS-1420: Schema page updates

evisdrenova:


frontend/apps/web/components/jobs/SchemaTable/FormErrorsCard.tsx
```diff
@@ -35,8 +35,8 @@ export default function FormErrorsCard(props: Props): ReactElement {
   const messages = formErrorsToMessages(formErrors);
   return (
     <Card className="w-full flex flex-col">
-      <CardHeader className="flex flex-col gap-2">
-        <div className="flex flex-row items-center justify-between">
+      <CardHeader className="flex flex-col">
+        <div className="flex flex-row items-center justify-between h-8">
```
nickzelei, line 39: I tested a few forms that use this component on staging and production and they overall look fine to me in that when they render, they don't adjust the components at all (which that other PR seems to be changing?). So why the change here?

Also, this form errors card still uses the simpler form error syntax that just renders the text as red.
Did the changes to the `<FormMessage />` component bring on the need to adjust the height? Or was that an issue with the original shadcn component styling too?

evisdrenova: it's an issue with the original component before any of my <FormMessage /> updates. The <FormMessage /> updates just reveled it. 

The original row without with badge is 16px tall.
image

When the badge gets added in, it jumps to 22px
image

That 6px causes things to jump down 6px to make room for it. 

So with this change:

We set the row height to 32px so that it can accommodate the badge:
image

https://www.loom.com/share/7fee6c5a9839412e87330682cec5046d

nickzelei (approved):

PR #2644, opened 2024-09-09
NEOS-1425 Fixes SQL Server unique identifier type in sync

alishakawaguchi:


internal/sqlserver/utils.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	return jObj, nil
+}
+
+func IsUuidDataType(colDataType string) bool {
+	return strings.EqualFold(colDataType, "uniqueidentifier")
+}
+
+func BitsToUuidString(bits []byte) (string, error) {
```
nickzelei, line 64: can't believe the sqlserver driver doesn't handle this. but then again, I'm honestly astounded at how all of the drivers seem to lack official support for many of the official DB types...

nickzelei (approved):
LGTM, nicely done!

PR #2650, opened 2024-09-10
NEOS-1417: update yup error messages

evisdrenova:


frontend/apps/web/app/(mgmt)/[account]/new/job/job-form-validations.ts
```diff
@@ -129,12 +140,14 @@ export type DefineFormValues = Yup.InferType<typeof DefineFormValues>;
 
 export const ConnectFormValues = SourceFormValues.concat(
   Yup.object({
-    destinations: Yup.array(NewDestinationFormValues).required(),
+    destinations: Yup.array(NewDestinationFormValues).required(
+      'At least one destination Connection is required'
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/new/job/job-form-validations.ts:144]: fwiw the `required` here simply says that the `destinations` property is required, not that it has a minimum length (we arent really concerned with that here). In this case it just updates the type to remove the undefined. 

This is generally true for most of the `required` functions (meaning: it's mostly for us to have nice types as we typically have defaults for everything due to how that interacts with react forms)

frontend/apps/web/app/api/webhooks/keycloak/slack/route.ts
```diff
@@ -62,18 +62,20 @@ Example register event for google
   }
 */
 const RegisterEvent = Yup.object({
```
nickzelei, line 64: A little overkill for webhooks, but it's appreciated nonetheless 😅

frontend/apps/web/yup-validations/connections.ts
```diff
@@ -76,23 +76,30 @@ const connectionNameSchema = Yup.string()
 export const SshTunnelFormValues = Yup.object({
   host: Yup.string(),
   port: Yup.number()
-    .min(0)
-    .when('host', (host, schema) => ([host] ? schema.required() : schema)),
+    .min(0, 'The Port must be greater than 0.')
```
nickzelei, line 79: not sure this message is quite right, we allow the minimum to be 0. So it should be greater than or equal to.

nickzelei (approved):
LGTM! Left some notes, and one message that should be greater than or equal to instead of just greater than.

frontend/apps/web/app/(mgmt)/[account]/new/job/job-form-validations.ts
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/new/job/job-form-validations.ts:144]: yeah i was kinda going overkill to just set it

PR #2653, opened 2024-09-11
NEOS-1439: update name field lengths

evisdrenova:
Updates the `name` yup validation regex on the frontend and the backend regex we have in protos to allow names for jobs, connections, transformers and API keys to be up to 100 characters long. We could do it longer I suppose but I felt like 100 was reasonable. 

This also adds in styling support for truncating the long names throughout the app. I updated the `<TruncatedText /> ` component to expose a few more props which are useful for positioning the tooltip and setting the maxWidth. 

I also made some updates to responsiveness in the headers to try and make the resizing smoother. 

Note: there is another PR:  #2650 which handles updating all of the error messages so I didn't update them in this one. 

https://www.loom.com/share/e32a374f44564e7caf270b6797f5582c?sid=86a63742-41fa-4782-ba3b-1f332dc15071


frontend/apps/web/components/TruncatedText.tsx
```diff
@@ -7,12 +7,15 @@ import {
   TooltipTrigger,
 } from './ui/tooltip';
 
+type TooltipSide = 'top' | 'right' | 'bottom' | 'left';
+type TooltipAlign = 'start' | 'center' | 'end';
 interface Props {
   text: string;
-
+  side?: TooltipSide;
+  align?: TooltipAlign;
```
nickzelei, line 15: Instead of creating your own TS types, you can just use the ones provided by the tooltipcontent props.
```suggestion
  side?: TooltipContentProps['side']
  align?: TooltipContentProps['align']
```

frontend/apps/web/components/headers/PageHeader.tsx
```diff
@@ -33,21 +33,24 @@ export default function PageHeader(props: Props) {
...
           </div>
         </div>
-
-        {progressSteps && (
-          <div className="flex-1 flex justify-center">{progressSteps}</div>
-        )}
-        {extraHeading ? <div>{extraHeading}</div> : null}
+        <div className="flex-1 flex justify-center">
+          {progressSteps && progressSteps}
+        </div>
```
nickzelei, line 52: Hm, I think before we had it to only render the div if there were actually progress steps. Where as now this div will always be rendered regardless, which takes up some space.
Do we want to move it back to conditionally render the div only if there are steps?

frontend/apps/web/components/site-header/AccountSwitcher.tsx
```diff
@@ -54,7 +54,7 @@ const CreateTeamFormValues = Yup.object({
   name: Yup.string()
     .required()
     .min(3)
-    .max(30)
+    .max(100)
```
nickzelei, line 57: Should these have custom messages?

frontend/apps/web/yup-validations/connections.ts
```diff
@@ -77,7 +77,7 @@ export const SshTunnelFormValues = Yup.object({
   host: Yup.string(),
   port: Yup.number()
     .min(0)
-    .when('host', (host, schema) => ([host] ? schema.required() : schema)),
+    .when('host', (host, schema) => (host ? schema.required() : schema)),
```
nickzelei, line 80: Hm, this change seems a little suspect. What changed here that this should not be `[host]` anymore?
I suppose that made it always truthy so it seems like a bug, but how come we didnt catch this before?

nickzelei (approved):
left minor comments but otherwise LGTM.

Seems to be a bit more than just name field updates (lots of header style changes?) Curious what prompted those.

frontend/apps/web/components/ResourceId.tsx
```diff
@@ -20,7 +20,7 @@ export default function ResourceId(props: Props): ReactElement {
   } = props;
 
   return (
-    <div className="flex flex-col md:flex-row items-center gap-1">
+    <div className="flex flex-row items-center gap-1">
```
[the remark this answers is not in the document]
nickzelei: However maybe your other changes are making this just fine

frontend/apps/web/components/site-header/AccountSwitcher.tsx
nickzelei: Maybe your other PR handles that tho

nickzelei:
also thanks for providing the loom, was very helpful in seeing the style changes!

frontend/apps/web/components/ResourceId.tsx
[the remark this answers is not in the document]
evisdrenova: yeah responsive the changes i made account for it

frontend/apps/web/yup-validations/connections.ts
evisdrenova: yeah im not sure - js will always see that as truthy so it wasn't really doing anything which is why i removed it

PR #2655, opened 2024-09-11
NEOS-1441: adds dialog for showing SELECT from job run

nickzelei:
https://www.loom.com/share/43ef79052a3449ec829b51a18c419061

^ Demo

Fast Follows:

- Update Dialog to use Monaco editor
- When piping into Monaco, would be great if we could run the SQL through a formatter so that it looks nicer.
- Add loading state when clicking on View button

This is an 80/20 PR to unblock folks that want to access the select queries.

### DIFF
--- a/backend/services/mgmt/v1alpha1/job-service/runs.go
+++ b/backend/services/mgmt/v1alpha1/job-service/runs.go
@@ -677,9 +677,6 @@ func (s *Service) GetRunContext(
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


evisdrenova (approved):
Overall I think looks good for a v0! Could be nice to style the code using a <pre> tag but that can also be a fast follow

alishakawaguchi (approved):

nickzelei:
> Overall I think looks good for a v0! Could be nice to style the code using a
> 
>  tag but that can also be a fast follow

yep, I listed the fast follows in the description.
will be creating tickets for those to work them later.

PR #2665, opened 2024-09-11
NEOS-1438 Destination option to skip foreign key constraint errors in sync

alishakawaguchi:


worker/pkg/benthos/sql/output_sql_insert.go
```diff
@@ -283,8 +292,41 @@ func (s *pooledInsertOutput) WriteBatch(ctx context.Context, batch service.Messa
...
+
+func (s *pooledInsertOutput) RetryInsertRowByRow(
+	ctx context.Context,
+	columns []string,
+	rows [][]any,
+) error {
+	errorCount := 0
+	insertCount := 0
+	for _, row := range rows {
+		insertQuery, err := querybuilder.BuildInsertQuery(s.driver, fmt.Sprintf("%s.%s", s.schema, s.table), columns, [][]any{row}, &s.onConflictDoNothing)
```
nickzelei, line 314: `fmt.Sprintf("%s.%s", s.schema, s.table)`
I know this gets fed into goqu and properly escaped, but what if the schema or table has a period in it? I'm guessing it doesnt know what to do. Anyways, idk if we are doing this anywhere but now that I'm hyper aware of special chars thought i'd point this out...

nickzelei: Maybe in a follow up we should update those `BuildInsertQuery` and `BuildUpdateQuery` functions to call out schema and table as specific parameters so that we can use goqu to build them like `goqu.T(table).Schema(schema)` so that it can handle the escaping for us.

nickzelei (approved):
LGTM, left a minor comment about a follow up we should do.

PR #2674, opened 2024-09-12
NEOS-1449: adds a running timer to the run id page

evisdrenova:
This feature adds a running timer to the duration stat card in the run/[id] page. Gives users an idea of how long a job is taking. I noticed this when I was working with a customer that they were looking at the timer as a way of seeing if they were getting past the issue i.e. the last job failed at 25 seconds, now that we're at 40 seconds, we likely resolved that. 

The timer is pretty close but not perfect, in that it usually goes over by 2-3 seconds but then updates to the final time once it's there. That's also just locally, in prod it likely is better. 

This handles the minutes and seconds conversions as well so if it goes over 59 seconds it will change from '59 seconds' to '1 minute 1 second'. It also handles plural seconds and minutes. It does not go into hours. Can make that a fast follow. 

https://www.loom.com/share/f14d4c15bf244fd59c67c915c8b6ab7d

### DIFF
--- a/compose.dev.yml
+++ b/compose.dev.yml
@@ -6,7 +6,7 @@ include:
       - ./compose/temporal/.env
   - path: ./compose/compose-db.yml
   # - path: ./compose/compose-db-mysql.yml
-  - path: ./compose/compose-db-mongo.yml
+  # - path: ./compose/compose-db-mongo.yml
   # - path: ./compose/compose-db-dynamo.yml
   # - path: ./compose/compose-db-mssql.yml
 
--- a/compose.dev.yml
+++ b/compose.dev.yml
@@ -6,7 +6,7 @@ include:
       - ./compose/temporal/.env
   - path: ./compose/compose-db.yml
   # - path: ./compose/compose-db-mysql.yml
-  # - path: ./compose/compose-db-mongo.yml
+  - path: ./compose/compose-db-mongo.yml
   # - path: ./compose/compose-db-dynamo.yml
   # - path: ./compose/compose-db-mssql.yml
 
--- a/frontend/apps/web/app/(mgmt)/[account]/runs/[id]/page.tsx
+++ b/frontend/apps/web/app/(mgmt)/[account]/runs/[id]/page.tsx
@@ -43,7 +43,7 @@ import {
 } from '@neosync/sdk/connectquery';
 import { ArrowRightIcon, Cross2Icon, TrashIcon } from '@radix-ui/react-icons';
 import { useRouter } from 'next/navigation';
-import { ReactElement, useState } from 'react';
+import { ReactElement, useEffect, useState } from 'react';
 import { toast } from 'sonner';
 import yaml from 'yaml';
 import JobRunStatus from '../components/JobRunStatus';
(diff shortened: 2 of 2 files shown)


frontend/apps/web/app/(mgmt)/[account]/runs/[id]/page.tsx
```diff
@@ -108,6 +108,43 @@ export default function Page({ params }: PageProps): ReactElement {
...
+          );
+          const minutes = Math.floor(diffInSeconds / 60);
+          const seconds = diffInSeconds % 60;
+          // handle plural minutes and seconds and set the duration
+          setDuration(
+            minutes > 0
+              ? `${minutes} minute${minutes > 1 ? 's' : ''} ${seconds} second${seconds !== 1 ? 's' : ''}`
+              : `${seconds} second${seconds !== 1 ? 's' : ''}`
+          );
+        }
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/runs/[id]/page.tsx:131]: I think you got the logic right in the else branch. `setDuration(getDuration(end, start))`
Isn't the end in this case `const now = new Date()` ?
Not sure there is really a need to duplicate the logic that is already written in the `getDuration` method.

frontend/apps/web/app/(mgmt)/[account]/runs/[id]/page.tsx
```diff
@@ -435,18 +466,18 @@ function StatCard(props: StatCardProps): ReactElement {
...
+  var diff = (completedAt.getTime() - startedAt.getTime()) / 1000;
+  const minutes = Math.floor(diff / 60);
+  const seconds = Math.round(diff % 60);
+  // handle plural minutes and seconds
   if (minutes === 0) {
-    return `${seconds} seconds`;
+    return `${seconds} second${seconds !== 1 ? 's' : ''}`;
   }
-  return `${minutes} minutes ${seconds} seconds`;
+  return `${minutes} minute${minutes > 1 ? 's' : ''} ${seconds} second${seconds !== 1 ? 's' : ''}`;
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/runs/[id]/page.tsx:480]: Could we replace all of this logic with [formatDuration](https://date-fns.org/v3.6.0/docs/formatDuration) ? I think that does exactly what we want here, but better.

nickzelei (commented):
this is pretty cool! Left a few comments on cleaning up the logic a bit.

I think we can use date-fns to make this even cleaner.

frontend/apps/web/app/(mgmt)/[account]/runs/[id]/page.tsx
```diff
@@ -108,6 +108,43 @@ export default function Page({ params }: PageProps): ReactElement {
...
+              ? `${minutes} minute${minutes > 1 ? 's' : ''} ${seconds} second${seconds !== 1 ? 's' : ''}`
+              : `${seconds} second${seconds !== 1 ? 's' : ''}`
+          );
+        }
+      };
+
+      updateDuration();
+      // sets up an interval to call the timer every second
+      timer = setInterval(updateDuration, 1000);
+    } else if (jobRun?.completedAt && jobRun?.startedAt) {
```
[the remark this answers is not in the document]
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/runs/[id]/page.tsx:137]: it does that on line 143 in the `return`, that always runs regardless of the job status

frontend/apps/web/app/(mgmt)/[account]/runs/[id]/page.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/runs/[id]/page.tsx:131]: makes sense, will clean it up

compose.dev.yml
```diff
@@ -6,7 +6,7 @@ include:
       - ./compose/temporal/.env
   - path: ./compose/compose-db.yml
   # - path: ./compose/compose-db-mysql.yml
-  # - path: ./compose/compose-db-mongo.yml
+  - path: ./compose/compose-db-mongo.yml
```
nickzelei, line 9: please remove this :)

frontend/apps/web/app/(mgmt)/[account]/runs/[id]/page.tsx
```diff
@@ -108,6 +109,39 @@ export default function Page({ params }: PageProps): ReactElement {
...
+        const now = new Date();
+        const startTime = jobRun?.startedAt?.toDate();
+        if (startTime) {
+          const diffInSeconds = Math.floor(
+            (now.getTime() - startTime.getTime()) / 1000
+          );
+          const minutes = Math.floor(diffInSeconds / 60);
+          const seconds = diffInSeconds % 60;
+          // handle plural minutes and seconds and set the duration
+          setDuration(formatDuration({ minutes: minutes, seconds: seconds }));
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/runs/[id]/page.tsx:127]: Why can't this just be:
```suggestion
          setDuration(getDuration(new Date(), startTime);
```

compose.dev.yml
evisdrenova: agh everytime ..

frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx
```diff
@@ -2,16 +2,16 @@ import { useAccount } from '@/components/providers/account-provider';
 import { Alert, AlertDescription } from '@/components/ui/alert';
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx:2]: Is this supposed to be in this PR?

Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx:2]: the alert import? yeah that's used to tell the user that there is no config options for a transformer online 207

Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx:2]: hm weird not sure why it got included. Will clean up.

nickzelei (approved):

PR #2690, opened 2024-09-13
NEOS-1462: Add transformer links to app

evisdrenova:
This features adds a "Learn More" link that points to the specific Transformer being inspected/selected. 

Updated:
1. System Transformers [id] page
2. Custom Transformers [id] page
3. Create new Custom Transformer page
4. Transformer Dialog in Mapping page

https://www.loom.com/share/632ff6d558954d86a411166f679f2659?sid=c9c5dc70-295c-45c8-ab5d-0b2cc788899d

### DIFF
--- a/frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx
+++ b/frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx
@@ -2,6 +2,7 @@
 
 import OverviewContainer from '@/components/containers/OverviewContainer';
 import PageHeader from '@/components/headers/PageHeader';
+import LearnMoreLink from '@/components/labels/LearnMoreLink';
 import { useAccount } from '@/components/providers/account-provider';
 import { Button } from '@/components/ui/button';
 import {
@@ -53,6 +54,7 @@ import {
   CreateUserDefinedTransformerFormContext,
   CreateUserDefinedTransformerFormValues,
 } from '../../../../../yup-validations/transformer-validations';
+import { constructDocsLink } from '../../transformers/EditTransformerOptions';
 import TransformerForm from './TransformerForms/TransformerForm';
 
 function getTransformerSource(sourceStr: string): TransformerSource {
@@ -180,7 +182,17 @@ export default function NewTransformer(): ReactElement {
               <FormItem>
                 <FormLabel>Source Transformer</FormLabel>
                 <FormDescription>
-                  The system transformer to clone.
+                  The system transformer to clone.{' '}
+                  {formSource !== 0 &&
+                    formSource !== null &&
+                    TransformerSource[formSource ?? 0] !=
+                      'GENERATE_JAVASCRIPT' &&
+                    TransformerSource[formSource ?? 0] !=
+                      'TRANSFORM_JAVASCRIPT' && (
+                      <LearnMoreLink
+                        href={constructDocsLink(formSource ?? 0)}
+                      />
+                    )}
                 </FormDescription>
                 <FormControl>
                   <Select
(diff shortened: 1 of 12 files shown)


frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx
```diff
@@ -180,7 +182,17 @@ export default function NewTransformer(): ReactElement {
...
                 <FormLabel>Source Transformer</FormLabel>
                 <FormDescription>
-                  The system transformer to clone.
+                  The system transformer to clone.{' '}
+                  {formSource !== 0 &&
+                    formSource !== null &&
+                    TransformerSource[formSource ?? 0] !=
+                      'GENERATE_JAVASCRIPT' &&
+                    TransformerSource[formSource ?? 0] !=
+                      'TRANSFORM_JAVASCRIPT' && (
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx:191]: There is a function that can be called  in this file (`getTransformerSource`) that will convert the number to `TransformerSource`. That way you can use typesafe values.

Comment (nickzelei) [reply frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx:191]: Also, why is ti being omitted for the javascript ones? Is there not a doc page for those?

frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx
```diff
@@ -216,3 +232,11 @@ function NoAdditionalTransformerConfigurations(): ReactElement {
     </Alert>
   );
 }
+
+export function constructDocsLink(source: number): string {
+  const name: string = TransformerSource[source]
+    .toLowerCase()
+    .replaceAll('_', '-');
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx:239]: we do have this function in the utils `getTransformerSourceString` that effectively does this (minute the replace all). Not sure if that is useful here.

frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx
```diff
@@ -93,7 +95,21 @@ export default function EditTransformerOptions(props: Props): ReactElement {
...
-              <DialogDescription>{transformer?.description}</DialogDescription>
+              <div className="flex flex-row items-center gap-2">
+                <DialogDescription>
+                  {transformer?.description}{' '}
+                  {/* don't render the learn more link for the js transformers since they have their own special link && only render for system transformers*/}
+                  {TransformerSource[transformer.source] !=
+                    'GENERATE_JAVASCRIPT' &&
+                    TransformerSource[transformer.source] !=
+                      'TRANSFORM_JAVASCRIPT' &&
+                    isSystemTransformer(transformer) && (
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx:106]: would be great if this could be pulled into a function or something since it's now used in a few spots.

Also, when comparing the transformer sources, you can just use the importable enumeration instead of doing string comparisons.
`transformer.source !== TransformerSource.GENERATE_JAVASCRIPT` etc.

nickzelei (changes requested):
Thx for adding these.

Two main things:

Instead of doing the string comparisons, use the `TransformerSource` enum from the `@neosync/sdk`.

Please pull the logic for determining if a transformer source / transformer should render a link into a function in the `util/util.ts` file so that it can be shared as right now it's copied in like 3-4 places.

Eventually it would be nice if the transformer is user defined if we could de-reference it into the underlying transformer source so that the learn more link can be rendered for those too.

frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/new/transformer/page.tsx:191]: Javascript ones have their own special docs link that points to the custom transformers page which is a guide vs. these docs link go to the system transformers. I was originally thinking of leaving it there since the learn more link was right next to the code input instead of having to scroll up to the top of the page. But i can update the `constructDocLink` function to handle that case.

frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/transformers/EditTransformerOptions.tsx:239]: yeah we need the replace since the link sin the docs has dashes while the transformer source has underscores

PR #2695, opened 2024-09-14
NEOS-1419: add real time connection validation for the Generate job

evisdrenova:
Adds in real time connection checking in the Generate job Connect page similar to the Sync job page. 

https://www.loom.com/share/e30eb0f57c53460f96732dc33bc5f495?sid=59d8ae18-09a9-46c6-811a-431839e0778e

### DIFF
--- a/frontend/apps/web/app/(mgmt)/[account]/new/job/generate/single/connect/page.tsx
+++ b/frontend/apps/web/app/(mgmt)/[account]/new/job/generate/single/connect/page.tsx
@@ -205,6 +205,12 @@ export default function Page({ searchParams }: PageProps): ReactElement {
                               const destId = form.getValues(
                                 'destination.connectionId'
                               );
+                              if (!destId) {
+                                form.setValue(
+                                  'destination.connectionId',
+                                  value
+                                );
+                              }
 
                               setIsSourceValidating(true);
                               try {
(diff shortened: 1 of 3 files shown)


frontend/apps/web/app/(mgmt)/[account]/new/job/generate/single/connect/page.tsx
```diff
@@ -145,71 +161,117 @@ export default function Page({ searchParams }: PageProps): ReactElement {
...
-                              );
-                              const connType = getConnectionType(
-                                connection?.connectionConfig ??
-                                  new ConnectionConfig()
-                              );
-                              if (connType) {
-                                urlParams.append('connectionType', connType);
+                              setIsSourceValidating(true);
+                              try {
+                                const res = await checkConnectionConfig({
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/new/job/generate/single/connect/page.tsx:211]: This is going to block the UI from updating the `destination.connectionId` until this returns, which could be bad if the connection is broken or some other reason.

In other words, this should be not awaited here and instead done as a standard promise, in a separate effect, or the below code should be moved to above this.

nickzelei (changes requested):
left one comment about the ui being blocked due to the validation

nickzelei (approved):

PR #2702, opened 2024-09-16
Reduces bastion host network error logs

nickzelei:
Updates a common (unavoidable) error to be a warning.
Adds more debug logging to bastion host.
Took some time to rework the host to be much easier to read and better handle connection leaks.

I've had to manually test this since we don't have reliable integration tests for tunneling yet.

### DIFF
--- a/backend/pkg/sshtunnel/tunnel.go
+++ b/backend/pkg/sshtunnel/tunnel.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/google/uuid"
 	"golang.org/x/crypto/ssh"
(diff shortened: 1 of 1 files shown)


backend/pkg/sshtunnel/tunnel.go
```diff
@@ -54,6 +56,7 @@ func New(
 			User:            tunnel.User,
 			Auth:            authmethods,
 			HostKeyCallback: getHostKeyCallback(serverPublicKey),
+			Timeout:         30 * time.Second,
```
nickzelei, line 59: Intentional. Without this it will effectively wait forever, which is not good.
NEOS-1470 to expose this as a configuration option for users

backend/pkg/sshtunnel/tunnel.go
```diff
@@ -85,45 +88,54 @@ func (t *Sshtunnel) Start(logger *slog.Logger) (chan any, error) {
 }
 
 func (t *Sshtunnel) serve(listener net.Listener, ready chan<- any, logger *slog.Logger) {
+	defer func() {
```
nickzelei, line 91: Ensures we always close the listener now.

alishakawaguchi (approved):

PR #2714, opened 2024-09-18
feat(api-chart): allow setting extraEnvVars from values in initContainer

WhiteDiamondz:
In my use case I wished to set sensitive values like DB_PASS as environment variables loaded from secrets.

The user-container of the api makes this posible by using the values the user has defined in .Values.extraEnvVars but the same does not exist for the init-container that still needs similar sensitive value to perform the migration tasks.

This pull request is a suggestion that would fix this problem and allow setting environment variables in this way to override the ones loaded in `db-migrations-env-vars.yaml` that rely on values directly set in clear.

```
extraEnvVars:
  - name: DB_PASS
    valueFrom:
      secretKeyRef:
        name: neosync-secret
        key: DB_PASS
```

### DIFF
--- a/backend/charts/api/templates/deployment.yaml
+++ b/backend/charts/api/templates/deployment.yaml
@@ -83,6 +83,33 @@ spec:
             - migrate
             - up
 
+          env:
+
+            {{- with .Values.extraEnvVars }}
+            {{- range .}}
+            - name: {{ .name }}
+              {{- if .value }}
+              value: {{ .value | quote }}
+              {{- else if .valueFrom }}
+              valueFrom:
+                {{- if .valueFrom.secretKeyRef }}
+                secretKeyRef:
+                  name: {{ .valueFrom.secretKeyRef.name }}
+                  key: {{ .valueFrom.secretKeyRef.key }}
+                {{- end }}
+                {{- if .valueFrom.configMapKeyRef }}
+                configMapKeyRef:
+                  name: {{ .valueFrom.configMapKeyRef.name }}
+                  key: {{ .valueFrom.configMapKeyRef.key }}
+                {{- end }}
+                {{- if .valueFrom.fieldRef }}
+                fieldRef:
+                  fieldPath: {{ .valueFrom.fieldRef.fieldPath }}
+                {{- end }}
+              {{- end }}
+            {{- end }}
+            {{- end }}
+
           envFrom:
             - secretRef:
                 name: {{ template "neosync-api.fullname" . }}-migration-evs


WhiteDiamondz:
Just realised that a [similar PR](https://github.com/nucleuscloud/neosync/pull/2602) had been opened a few weeks ago. Perhaps my context helps to understand the need behind those PRs. In my case I did not add a new section to the values under the migrations but this could be done in the case we would want to pass the same environment variable with different values between the migration init container and the user-container.

Feel free to close as duplicate if you wish to continue the disucssion on the other one !

nickzelei:
Hey @WhiteDiamondz , appreciate the PR!

I checked out what you changed and I think the main difference between your PR and #2602 is that they have `.Values.migrations.extraEnvVars` which is slightly different from this one which is just `.Values.extraEnvVars`

I'm unclear if this is a good practice or not or if it would surprise folks that env vars from this property get applied to both places (maybe it's already not clear that the top level extraEnvVars gets applied to the user container at all?)

Appreciate your thoughts here, and I also appreciate you explaining why it's of interest!

We havent had the need for this yet because the helm chart stores all of the helm values in a secret anyways, but it seems like you are trying to avoid re-defining the same value in multiple places (in our case the migrations db user is separate from the main user-container).

nickzelei:
Hey @WhiteDiamondz so I ended up cleaning and merging in a version that uses the `.Values.migrations.extraEnvVars` - I hope that gets you where you need to go! I'm going to close this PR, but if you think there is still a need or value for adding the top level `extraEnvVars` to the migrations, feel free to reach out to me in our Discord, or open another issue/PR and we can talk about it.

WhiteDiamondz:
Hey @nickzelei !

Thanks for the quick reply !
Indeed, the main difference was the use of the `.Values.extraEnvVars` for both containers. I admit I was a bit thrown off at first, I didn't expect this to only be used in the user container and not the init-container and that is why I was suggesting this change. 

I don't know much about database good practices but in my case I didn't really see a reason on why I should use a different account (DB_USER, DB_PASS) between `.Values.db` and `.Values.migrations.db`. Thus, using common variables seemed convenient but that may not be a good practice (?

I probably should have insisted more on the specific pain point that motivated this Push Request in my initial explanation. 
Even if the helm chart **does** stores all of the helm values in a secret, it still forces you to specify sensible values like the database password in the values. Since we push these files to our git it was not an acceptable solution (even if they would have been in a secret on our Kube cluster). 

I was able to set the value by directly using the `.Values.extraEnvVars` from a secret that fetches it's value from a Hashicorp Vault instance (so no risk of pushing sensible data).

```
extraEnvVars:
  - name: DB_PASS
    valueFrom:
      secretKeyRef:
        name: neosync-secret-created-by-vault-operator
        key: DB_PASS
```

However, this wasn't able to work for the db migrations since I had no way to override the environment variables by fetching them from an outside secret. Adding `.Values.migrations.extraEnvVars` is a solution that will make this a lot easier and solves my problem ! I will just need to replicate what I had done for the user-container.

Thank you for merging this fix !

nickzelei:
> Hey @nickzelei !
> 
> Thanks for the quick reply ! Indeed, the main difference was the use of the `.Values.extraEnvVars` for both containers. I admit I was a bit thrown off at first, I didn't expect this to only be used in the user container and not the init-container and that is why I was suggesting this change.
> 
> I don't know much about database good practices but in my case I didn't really see a reason on why I should use a different account (DB_USER, DB_PASS) between `.Values.db` and `.Values.migrations.db`. Thus, using common variables seemed convenient but that may not be a good practice (?
> 
> I probably should have insisted more on the specific pain point that motivated this Push Request in my initial explanation. Even if the helm chart **does** stores all of the helm values in a secret, it still forces you to specify sensible values like the database password in the values. Since we push these files to our git it was not an acceptable solution (even if they would have been in a secret on our Kube cluster).
> 
> I was able to set the value by directly using the `.Values.extraEnvVars` from a secret that fetches it's value from a Hashicorp Vault instance (so no risk of pushing sensible data).
> 
> ```
> extraEnvVars:
>   - name: DB_PASS
>     valueFrom:
>       secretKeyRef:
>         name: neosync-secret-created-by-vault-operator
>         key: DB_PASS
> ```
> 
> However, this wasn't able to work for the db migrations since I had no way to override the environment variables by fetching them from an outside secret. Adding `.Values.migrations.extraEnvVars` is a solution that will make this a lot easier and solves my problem ! I will just need to replicate what I had done for the user-container.
> 
> Thank you for merging this fix !

Thanks for providing more color here @WhiteDiamondz - makes sense. It works for our setup because we use helm secrets for our gitops flow so all of our secret files are encrypted. But everyone has a different setup!

As far as best practices go for database users, it can be a good practice to have different users do these tasks.
For instance, in our setup, the user container database user only has access to read/write to the database tables. It does not have the ALTER capabilities and as such, cannot run schema migrations. The job of the user-container is to CRUD data to/fro those tables.

We have a different schema-level user that the init-container uses to run the database migrations with elevated permissions. This can be good to do from a least-principal mindset, but can be more complicated to initially setup.

WhiteDiamondz:
Thanks for providing more context as well !

The current solution seems to be perfect then. Congrats for the great tool you are working on ! :rocket:

PR #2715, opened 2024-09-18
Fix(worker-chart): remove empty label if value not set

WhiteDiamondz:
An empty label can cause issues in the deployment.

A solution could be setting a default value but I could not think of one that wouldn't interfer with the user. This is another posibility.


nickzelei (approved):
Appreciate this change! We should probably rename or remove the `nucleusEnv` at some point to just be `env` or something like that.

Definitely makes sense to have this be optional here.

PR #2720, opened 2024-09-19
NEOS-1249: add support for client tls to mysql

nickzelei:
was working on this but paused it for now until I can sink more time into figuring out how to ad-hoc apply client-tls certificates instead of registering them at the driver level.


nickzelei:
Closing this as stale. This will be worked via: NEOS-1504 (#2983) using dialers vs the old way we were handling it with URL params.

PR #2721, opened 2024-09-19
NEOS-1429, NEOS-1475: [MIGRATION]: adds self-checkout billing to neosync cloud

nickzelei:


backend/internal/nucleusdb/users_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
```
nickzelei, line 4: These are due for a rewrite to use a postgres test container.

backend/protos/mgmt/v1alpha1/user_account.proto
```diff
@@ -178,7 +195,7 @@ message GetAccountStatusResponse {
   optional uint64 allowed_record_count = 2;
 
   // The current subscription status determined by the billing system.
-  int32 subscription_status = 3;
+  BillingStatus subscription_status = 3;
```
nickzelei, line 198: this techncially breaks the api but it's still an int so it is fine (and unused...)

alishakawaguchi (approved):

PR #2728, opened 2024-09-21
NEOS-1483 SQL types to Javascript types transformer docs

alishakawaguchi:
Fixes xml types. converts from byte array to string


nickzelei (approved):
overall looks good but am curious as to why the proto docs changed in this PR

PR #2735, opened 2024-09-23
NEOS-1433 Updates javascript transformer to use patch function

alishakawaguchi:
This makes sure that only columns that have JS transformers set are updated. Prevents unnecessary conversion of all column values from Javascript back to golang

### DIFF
--- a/worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors.go
+++ b/worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors.go
@@ -368,10 +368,9 @@ func constructBenthosJsProcessor(jsFunctions, benthosOutputs []string) string {
 (() => {
 %s
 const input = benthos.v0_msg_as_structured();
-const output = { ...input };
 const updatedValues = {}
 %s
-neosync.patchMessage(updatedValues)
+neosync.patchStructuredMessage(updatedValues)
 })();`, jsFunctionStrings, benthosOutputString)
 	return jsCode
 }
--- a/worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors.go
+++ b/worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors.go
@@ -366,21 +366,12 @@ func constructBenthosJsProcessor(jsFunctions, benthosOutputs []string) string {
 
 	jsCode := fmt.Sprintf(`
 (() => {
-function setNestedProperty(obj, path, value) {
-	path.split('.').reduce((current, part, index, parts) => {
-		if (index === parts.length - 1) {
-			current[part] = value;
-		} else {
-			current[part] = current[part] ?? {};
-		}
-		return current[part];
-	}, obj);
-}
 %s
 const input = benthos.v0_msg_as_structured();
 const output = { ...input };
+const updatedValues = {}
 %s
-benthos.v0_msg_set_structured(output);
+neosync.patchMessage(updatedValues)
 })();`, jsFunctionStrings, benthosOutputString)
 	return jsCode
 }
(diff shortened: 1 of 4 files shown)


worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors.go
```diff
@@ -366,21 +366,12 @@ func constructBenthosJsProcessor(jsFunctions, benthosOutputs []string) string {
 
 	jsCode := fmt.Sprintf(`
 (() => {
-function setNestedProperty(obj, path, value) {
```
nickzelei, line 369: 🙌

worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors.go
```diff
@@ -366,21 +366,12 @@ func constructBenthosJsProcessor(jsFunctions, benthosOutputs []string) string {
...
-			current[part] = value;
-		} else {
-			current[part] = current[part] ?? {};
-		}
-		return current[part];
-	}, obj);
-}
 %s
 const input = benthos.v0_msg_as_structured();
 const output = { ...input };
```
nickzelei, line 371: Is the output variable used anymore then?
Does it make sense to rename `updatedValues` to `output`? `patchedOutput`? `updatedValues`? `patchedValues`?

worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors.go
```diff
@@ -366,21 +366,12 @@ func constructBenthosJsProcessor(jsFunctions, benthosOutputs []string) string {
...
-		if (index === parts.length - 1) {
-			current[part] = value;
-		} else {
-			current[part] = current[part] ?? {};
-		}
-		return current[part];
-	}, obj);
-}
 %s
 const input = benthos.v0_msg_as_structured();
```
nickzelei, line 370: Just checking but we still need the `input` right because we pass it in to the transformer functions.

worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors_test.go
```diff
@@ -335,71 +343,9 @@ func Test_buildProcessorConfigsJavascript_DeepKeys(t *testing.T) {
 let programOutput = undefined;
 const benthos = {
   v0_msg_as_structured: () => ({foo: {bar: {baz: "world"}}}),
-  v0_msg_set_structured: (val) => {
```
nickzelei, line 338: Hard to tell based on the diff, but did these two tests just get combined?

worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors_test.go
```diff
@@ -276,7 +282,9 @@ func Test_buildProcessorConfigsTransformAndGenerateJavascript(t *testing.T) {
 let programOutput = undefined;
 const benthos = {
   v0_msg_as_structured: () => ({"name": "world"}),
-  v0_msg_set_structured: (val) => {
+};
+const neosync = {
+  patchMessage: (val) => {
```
nickzelei, line 287: I'm gonna nitpick this but should we rename this to something else that insinuates that it is expecting a structured message, similar to how benthos is doing their `msg_set_structured`? like `patchStructuredMessage`? or something? Or is this too much?  Just wondering because it _is_ a little different than what the message itself might be? I'm guessing there is a reason they have the "structured" suffix in there.
Anyways, lmk what you think there. `patchStructuredMessage` seems like a mouthful so maybe we just leave it as `patchMessage`, just wanted to put my thought down here.

nickzelei (approved):
Left a few minor comments/nits. Nothing that I think requires a change. Overall, looks good and nicely done!! 🚢

worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors.go
alishakawaguchi: yep still need the input

worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors_test.go
alishakawaguchi: yeah it became a duplicate test since I moved the setNestedProperty logic out of the javascript

worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors_test.go
alishakawaguchi: renamed to patchStructuredMessage 👍

PR #2740, opened 2024-09-24
Updates View Select Query to use Monaco Editor

nickzelei:
Also adds in a query formatted to display nicely.

https://www.loom.com/share/4d23df2e2370491baa476ea3d039a656

image

### DIFF
--- a/frontend/apps/web/package.json
+++ b/frontend/apps/web/package.json
@@ -65,6 +65,7 @@
     "react-icons": "^5.2.1",
     "recharts": "^2.12.7",
     "sonner": "^1.5.0",
+    "sql-formatter": "^15.4.2",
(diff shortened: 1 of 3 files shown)


frontend/apps/web/package.json
```diff
@@ -65,6 +65,7 @@
     "react-icons": "^5.2.1",
     "recharts": "^2.12.7",
     "sonner": "^1.5.0",
+    "sql-formatter": "^15.4.2",
```
nickzelei, line 68: One caveat here is this is platform specific. So if using docker dev, let docker do `npm install` otherwise the runs page will not load.

alishakawaguchi (approved):

evisdrenova (approved):

PR #2741, opened 2024-09-25
NEOS-1489 Fix default being converted to database DEFAULT

alishakawaguchi:


worker/pkg/workflows/datasync/activities/gen-benthos-configs/benthos-builder_test.go
```diff
@@ -148,8 +183,10 @@ func Test_ProcessorConfigEmptyJavascript(t *testing.T) {
 	queryMap := map[string]map[tabledependency.RunType]string{
 		"public.users": {tabledependency.RunTypeInsert: ""},
 	}
+	logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
```
nickzelei, line 186: for the future, if you don't want this to log during the tests you can pass in `io.Discard` instead of `os.Stdout`.
You can search the codebase for `discardLogger` to see examples that I've been creating.

nickzelei: I think at some point we probably want to have a "test logger" that we can pass in and maybe enable it if we use the `-v` flag for tests or something. Would be a nice QoL

nickzelei (approved):

PR #2743, opened 2024-09-25
NEOS-1484 Reset SQL column defaults after sync

alishakawaguchi:


backend/pkg/sqlmanager/postgres/postgres-manager.go
```diff
@@ -875,7 +875,7 @@ func (p *PostgresManager) GetTableRowCount(
 func BuildPgTruncateStatement(
 	tables []string,
 ) string {
-	return fmt.Sprintf("TRUNCATE TABLE %s;", strings.Join(tables, ", "))
+	return fmt.Sprintf("TRUNCATE TABLE %s RESTART IDENTITY;", strings.Join(tables, ", "))
```
nickzelei, line 878: small side note, but interesting that this one builds it manually, but the cascade statement builds it using goqu 😄

worker/pkg/workflows/datasync/activities/post-table-sync/activity.go
```diff
@@ -0,0 +1,133 @@
...
+type RunPostTableSyncResponse struct {
+}
+
+func (a *Activity) RunPostTableSync(
+	ctx context.Context,
+	req *RunPostTableSyncRequest,
+) (*RunPostTableSyncResponse, error) {
+	activityInfo := activity.GetInfo(ctx)
+	loggerKeyVals := []any{
+		"jobId", req.JobId,
```
nickzelei, line 52: prob worth adding the accountId in here since it is a part of the request object.

worker/pkg/workflows/datasync/activities/post-table-sync/activity.go
```diff
@@ -0,0 +1,133 @@
...
+		"RunID", activityInfo.WorkflowExecution.RunID,
+	}
+	logger := log.With(
+		activity.GetLogger(ctx),
+		loggerKeyVals...,
+	)
+	logger.Debug("running post table sync activity")
+	slogger := neosynclogger.NewJsonSLogger().With(loggerKeyVals...)
+
+	jobResp, err := a.jobclient.GetJob(ctx, connect.NewRequest(&mgmtv1alpha1.GetJobRequest{Id: req.JobId}))
```
nickzelei, line 63: just curious but since you are retrieving the job here, do you need to pass in the account id to the request? You can probably get away with not having it since you are retrieving the job anyways.

worker/pkg/workflows/datasync/activities/post-table-sync/activity.go
```diff
@@ -0,0 +1,133 @@
...
+	req *RunPostTableSyncRequest,
+	slogger *slog.Logger,
+	job *mgmtv1alpha1.Job,
+	workflowId string,
+) error {
+	rcResp, err := a.jobclient.GetRunContext(ctx, connect.NewRequest(&mgmtv1alpha1.GetRunContextRequest{
+		Id: &mgmtv1alpha1.RunContextKey{
+			JobRunId:   workflowId,
+			ExternalId: shared.GetPostTableSyncConfigExternalId(req.Name),
+			AccountId:  req.AccountId,
```
nickzelei, line 92: You have the mgmt job so you could just pull that.

worker/pkg/workflows/datasync/activities/post-table-sync/activity.go
```diff
@@ -0,0 +1,133 @@
...
+	logger.Debug("running post table sync activity")
+	slogger := neosynclogger.NewJsonSLogger().With(loggerKeyVals...)
+
+	jobResp, err := a.jobclient.GetJob(ctx, connect.NewRequest(&mgmtv1alpha1.GetJobRequest{Id: req.JobId}))
+	if err != nil {
+		return nil, fmt.Errorf("unable to get job by id: %w", err)
+	}
+	job := jobResp.Msg.GetJob()
+
+	switch job.Source.Options.Config.(type) {
```
nickzelei, line 69: Prefer using the getters here. `job.GetSource().GetOptions()`  etc.

worker/pkg/workflows/datasync/activities/post-table-sync/activity.go
```diff
@@ -0,0 +1,133 @@
...
+			AccountId:  req.AccountId,
+		},
+	}))
+	if err != nil {
+		return fmt.Errorf("unable to retrieve posttablesync runcontext for %s: %w", req.Name, err)
+	}
+
+	configBits := rcResp.Msg.GetValue()
+	if len(configBits) == 0 {
+		return nil
```
nickzelei, line 101: is this expected? if not, could be worth adding a warn here.

worker/pkg/workflows/datasync/activities/post-table-sync/activity.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	var config *shared.SqlPostTableSyncConfig
+	err = json.Unmarshal(configBits, &config)
+	if err != nil {
+		return err
```
nickzelei, line 107: Might be worth wrapping this with a more informative error statement so that we know exactly. Doesnt have to be wild, just a wrapped here that states we weren't able to unmarshal that object of type. Mostly because I cant remember off the top of my head what the standard error response is from the json.Unmarshal, but something tells me its obtuse.

worker/pkg/workflows/datasync/activities/post-table-sync/activity.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	if len(config.DestinationStatements) == 0 {
+		return nil
```
nickzelei, line 110: Could be a good spot here to add a debug statement simply outputting that we exited early due to there being no destination statements. Or perhaps a more general debug log stating the number of destination statements we found.

worker/pkg/workflows/datasync/activities/post-table-sync/activity.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	for _, destination := range job.Destinations {
+		destinationConnection, err := shared.GetConnectionById(ctx, a.connclient, destination.ConnectionId)
+		if err != nil {
+			return fmt.Errorf("unable to get destination connection (%s) by id: %w", destination.ConnectionId, err)
+		}
+		destDb, err := a.sqlmanagerclient.NewPooledSqlDb(ctx, slogger, destinationConnection)
```
nickzelei, line 118: Is it safe to just call this here? What if the destination is not a sql-enabled connection? Or would it just not ever get this far?

worker/pkg/workflows/datasync/activities/post-table-sync/activity.go
```diff
@@ -0,0 +1,133 @@
...
+		if err != nil {
+			destDb.Db.Close()
+			slogger.Error("unable to connection to destination", "connectionId", destination.ConnectionId)
+			continue
+		}
+		for _, stmt := range config.DestinationStatements {
```
nickzelei, line 124: I'm a little confused by this logic here.
We iterate over the list of job destination connections - and we invoke the same set of destination statements for each one?

I guess the implicit assumption here is the following:
1. we only make it this far for SQL-enabled connections
2. each sql-enabled destination connection is of the same type (all PG, all MSSQL, all MYSQL) -- which is ok...because that is what we support today. no cross over.

I just read this code and it's not obvious, unless you know that to be true about Neosync, which is maybe the case in many places in our codebase, yet here I am pointing it out for this particular snippet.

worker/pkg/workflows/datasync/workflow/workflow.go
```diff
@@ -319,6 +325,30 @@ func retrieveActivityOptions(
...
+	logger log.Logger,
+	actOptResp *syncactivityopts_activity.RetrieveActivityOptionsResponse,
+	jobId string,
+	name string,
+) error {
+	logger.Debug("executing post table sync activity")
+	var resp *posttablesync_activity.RunPostTableSyncResponse
+	var postTableSyncActivity *posttablesync_activity.Activity
+	err := workflow.ExecuteActivity(
+		workflow.WithActivityOptions(ctx, *actOptResp.SyncActivityOptions),
```
nickzelei, line 339: Is it expected here that we want to use the SyncActivityOptions for this activity?

worker/pkg/workflows/datasync/activities/run-sql-init-table-stmts/init-statement-builder.go
```diff
@@ -199,6 +200,37 @@ func (b *initStatementBuilder) RunSqlInitTableStatements(
 					return nil, fmt.Errorf("unable to exec ordered truncate statements: %w", err)
 				}
 			}
+			if !sqlopts.InitSchema && (sqlopts.TruncateBeforeInsert || sqlopts.TruncateCascade) {
```
nickzelei, line 203: why are we only doing this if we are not also initializing the schema? The schema init is idempotent so what if the schema already exists? maybe I'm just missing some context here.

nickzelei (commented):
many questions. can address on here or chat in person tomorrow.

backend/pkg/sqlmanager/postgres/postgres-manager.go
alishakawaguchi: will update this as a fast follow

PR #2768, opened 2024-10-02
Use goqu to build postgres truncate statement

alishakawaguchi:


backend/pkg/sqlmanager/sql-manager.go
```diff
@@ -311,7 +311,7 @@ func (s *SqlManager) NewSqlDbFromUrl(
 ) (*SqlConnection, error) {
 	var db SqlDatabase
 	switch driver {
-	case sqlmanager_shared.PostgresDriver:
+	case sqlmanager_shared.PostgresDriver, "postgres":
```
alishakawaguchi, line 314: needed to support CLI sync

nickzelei (approved):

PR #2770, opened 2024-10-03
Updates Records Used Link to route to correct place

nickzelei:


frontend/apps/web/components/site-header/RecordsProgressBar.tsx
```diff
@@ -66,13 +68,22 @@ export default function RecordsProgressBar(props: Props): ReactElement {
...
+      }}
       variant="outline"
       className={cn(count > totalRecords && 'bg-orange-200 dark:bg-orange-500')}
     >
-      <div className="flex flex-row items-center gap-2 w-60">
-        <span className="text-sm text-nowrap">Records used </span>
-        <Progress value={percentageUsed} className="w-[60%]" />
+      <div className="flex flex-row items-center gap-2 sm:w-60">
+        <span className="text-sm text-nowrap">Records used</span>
+        <Progress value={percentageUsed} className="w-[60%] hidden sm:flex" />
```
nickzelei, line 86: Hides the progress bar on small screens.

frontend/apps/web/components/site-header/SiteHeader.tsx
```diff
@@ -31,7 +31,8 @@ export default function SiteHeader(): ReactElement {
         <MainNav />
         <MobileNav />
         <div className="flex flex-1 justify-end items-center space-x-2">
-          {systemAppConfig.isNeosyncCloud && <AccountsRecordProgress />}
+          {systemAppConfig.isNeosyncCloud &&
+            systemAppConfig.isStripeEnabled && <AccountsRecordProgress />}
```
nickzelei, line 35: This is better but not great. We should really be checking the account status and showing only if they have a max allowed records set. But this is good enough for now.

nickzelei: We can update the component to instead pull from the account status which has all of this information

evisdrenova (approved):

PR #2773, opened 2024-10-05
NEOS-1506:update empty states

evisdrenova:


worker/pkg/benthos/transformers/gen_generate_bool.go
```diff
@@ -39,7 +39,7 @@ func NewGenerateBoolOpts(
 func (t *GenerateBool) GetJsTemplateData() (*TemplateData, error) {
 	return &TemplateData{
 		Name: "generateBool",
-		Description: "Generates a boolean value at random.",
+		Description: "Generates a random boolean value.",
```
nickzelei, line 42: Hm, where are the source changes that caused all of the generated code to update? 
Or was it made in a past PR and we don't have the generation fully automated yet?
@alishakawaguchi

frontend/apps/web/app/(mgmt)/[account]/connections/page.tsx
```diff
@@ -59,7 +61,18 @@ function ConnectionTable(props: ConnectionTableProps): ReactElement {
 
   return (
     <div>
-      <DataTable columns={columns} data={connections} />
+      {connections.length == 0 ? (
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/connections/page.tsx:64]: just curious why we need this. We have a button on all of these pages that a user can click to create a new entity. Why do we need it in two spots?

Could we instead change the table's empty state to just say click the new button or something?

nickzelei (commented):
left a few comments.
Moreso questioning the need for the additional empty state/spot when all of these pages have a way already to create a new resource.

worker/pkg/benthos/transformers/gen_generate_bool.go
evisdrenova: These changes must have gotten picked up by accident - only meant to change the empty states. But in any case, the re-wording is valid. It aligns our docs with our descriptions in the benthos spec.

frontend/apps/web/app/(mgmt)/[account]/connections/page.tsx
Comment (evisdrenova) [reply frontend/apps/web/app/(mgmt)/[account]/connections/page.tsx:64]: Just puts the button in front of the user in the middle of the screen and gives them guidance on what to do. We could remove the button and just keep the title and description.

nickzelei (approved):

PR #2774, opened 2024-10-07
fix(mysql): correct default value handling for VARCHAR columns in Ini…

VincentRebs:
Hello,
When using the "Init Table Schema" option in Neosync to synchronize two MySQL 8.0.33 databases, a table creation error occurs if a VARCHAR column has a default value containing slashes (e.g., '/ranking/').
Error

In this example, the dir_path_dist column has a default value of '/ranking/'. When attempting to create the table using Neosync, an SQL error occurs. This error is caused by the default value being surrounded by parentheses, which is not allowed by MySQL for a VARCHAR default value.

**Source of the problem**
The issue lies in the file neosync/backend/pkg/sqlmanager/mysql/mysql-manager.go at line 356:
`if record.ColumnDefault != "" {
    pieces = append(pieces, fmt.Sprintf("DEFAULT (%s)", record.ColumnDefault));
}
`
This line wraps the default value in parentheses, which leads to an SQL syntax error when MySQL tries to create the table.

**Proposed Solution**
Replace the problematic line with the correct syntax that removes the parentheses and wraps the default value in single quotes:
`if record.ColumnDefault != "" {
    pieces = append(pieces, fmt.Sprintf("DEFAULT '%s'", record.ColumnDefault));
}
`
This change prevents the SQL syntax error when creating tables with default values.

After applying the modification, the table creation worked correctly without errors.

### DIFF
--- a/backend/pkg/sqlmanager/mysql/mysql-manager.go
+++ b/backend/pkg/sqlmanager/mysql/mysql-manager.go
@@ -15,6 +15,11 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const (
+	columnDefaultDefault = "Default"
+	columnDefaultString  = "String"
+)
+
 type MysqlManager struct {
 	querier mysql_queries.Querier
 	pool    mysql_queries.DBTX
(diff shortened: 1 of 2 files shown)


backend/pkg/sqlmanager/mysql/mysql-manager.go
```diff
@@ -350,7 +350,7 @@ func buildTableCol(record *buildTableColRequest) string {
 	}
 
 	if record.ColumnDefault != "" {
-		pieces = append(pieces, fmt.Sprintf("DEFAULT (%s)", record.ColumnDefault))
+		pieces = append(pieces, fmt.Sprintf("DEFAULT '%s'", record.ColumnDefault))
```
alishakawaguchi, line 353: Hi! Thanks for working on this bug. This definitely fixes default values with slashes but will break many other default values like CURRENT_TIMESTAMP and NOW() etc.. Need more logic here to handle these other cases

alishakawaguchi (changes requested):

backend/pkg/sqlmanager/mysql/mysql-manager.go
VincentRebs: Hi,
In this commit ( https://github.com/nucleuscloud/neosync/pull/2774/commits/34f020ae58533f3b105c6a301e4a46dbe49b9095 ) we tested several scenarios and noticed that the "extra" column always contains a value when the "default" column is a function. This value never appears when the "default" column is a string.

Based on this observation, we defined a new type called "ColumnDefaultType." It is set as a string when the column's default value contains information, and the "extra" column does not.

We then add quotes around the value if it’s a string, otherwise, we add parentheses as previously done.

Finally, if "ColumnDefaultType" is empty, nothing is done, just like before.

Tests were conducted on tables with both string default values and default values like now(). Everything works as expected.

This has resolved our issue while preserving the correct functionality when a function or other type is present.

Looking forward to your feedback.

alishakawaguchi (approved):
Looks good. Thanks for fixing!

alishakawaguchi:
@VincentRebs please fix lint error then I can merge it.

VincentRebs:
Hi,
The history has been cleared. I have fixed the lint errors.

PR #2783, opened 2024-10-08
NEOS-1333 AnonymizeMany and AnonymizeSingle endpoint

alishakawaguchi:


backend/protos/mgmt/v1alpha1/anonymization.proto
```diff
@@ -0,0 +1,65 @@
...
+
+import "buf/validate/validate.proto";
+import "mgmt/v1alpha1/transformer.proto";
+
+message AnonymizeManyRequest {
+  // Array of stringified JSON data to be anonymized (up to 25 items)
+  repeated string input_data = 1 [(buf.validate.field).repeated = {max_items: 25}];
+
+  // Transformer mappings using dot notation for field paths
+  map<string, TransformerConfig> transformer_mappings = 2;
```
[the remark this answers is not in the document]
nickzelei: What do you think? Map or slice?

internal/json-anonymizer/json-anonymizer.go
```diff
@@ -0,0 +1,369 @@
...
+	transformerExecutors       map[string]*transformer.TransformerExecutor
+	defaultTransformers        *mgmtv1alpha1.DefaultTransformersConfig
+	defaultTransformerExecutor *DefaultExecutors
+	compiledQuery              *gojq.Code
+	haltOnFailure              bool
+	skipPaths                  map[string]struct{}
+}
+
+// Option is a functional option for configuring the Anonymizer
+type Option func(*JsonAnonymizer)
```
nickzelei, line 33: Big fan of moving more of our classes to this setup.

internal/json-anonymizer/json-anonymizer.go
```diff
@@ -0,0 +1,369 @@
...
+	for rv.Kind() == reflect.Ptr {
+		if rv.IsNil() {
+			return nil
+		}
+		rv = rv.Elem()
+	}
+	return rv.Interface()
+}
+
+func customHash(input string) string {
```
nickzelei, line 148: Can this have a better name than just "customHash" ?

backend/services/mgmt/v1alpha1/anonymization-service/anonymization.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"context"
+
+	"connectrpc.com/connect"
+	mgmtv1alpha1 "github.com/nucleuscloud/neosync/backend/gen/go/protos/mgmt/v1alpha1"
+	jsonanonymizer "github.com/nucleuscloud/neosync/internal/json-anonymizer"
+)
+
+func (s *Service) AnonymizeMany(
```
nickzelei, line 11: Love how small and clean this file is.

backend/protos/mgmt/v1alpha1/anonymization.proto
```diff
@@ -0,0 +1,65 @@
...
+}
+
+message AnonymizeSingleResponse {
+  // Anonymized JSON data
+  string output_data = 1;
+}
+
+service AnonymizationService {
+  rpc AnonymizeMany(AnonymizeManyRequest) returns (AnonymizeManyResponse) {}
+  rpc AnonymizeSingle(AnonymizeSingleRequest) returns (AnonymizeSingleResponse) {}
```
nickzelei, line 64: Please add a small description comment to these methods.

nickzelei (changes requested):
Overall this is amazing. Super clean, nice work.
Left a few comments.

backend/protos/mgmt/v1alpha1/anonymization.proto
```diff
@@ -7,10 +7,16 @@ import "mgmt/v1alpha1/transformer.proto";
 
 message AnonymizeManyRequest {
   // Array of stringified JSON data to be anonymized (up to 25 items)
-  repeated string input_data = 1 [(buf.validate.field).repeated = {max_items: 25}];
+  repeated string input_data = 1 [(buf.validate.field).repeated = {
+    min_items: 1
+    max_items: 25
+    items: {
+      string: {len: 1}
```
nickzelei, line 14: is this right? For string primitives we defined it as `(buf.validate.field).string.min_len = 1`

this isn't requiring that everything is length 1? :)

backend/protos/mgmt/v1alpha1/anonymization.proto
```diff
@@ -27,8 +33,15 @@ message AnonymizeManyResponse {
   repeated AnonymizeManyErrors errors = 2;
 }
 
+message TransformerMapping {
+  // JQ Expression or Field Path to apply the transformation to
+  string expression = 1;
```
nickzelei, line 38: Should require that it is min length of 1. `(buf.validate.field).string.min_len = 1`

nickzelei: I see in the docs that there is also `[(buf.validate.field).required = true];` which according to the docs it says:
`// - proto3 scalar fields must be non-zero to be considered populated`
Idk if this has always been a thing or maybe at some point it didnt work which is why we use the min_len, but the .required seems a lot more descriptive.

backend/protos/mgmt/v1alpha1/anonymization.proto
```diff
@@ -27,8 +33,15 @@ message AnonymizeManyResponse {
...
+  // Configuration of Transformer to apply
+  TransformerConfig transformer = 2;
```
nickzelei, line 40: I suppose we could use the required field here too!

```

// If `required` is true, the field must be populated. A populated field can be
--
// described as "serialized in the wire format," which includes:
//
// - the following "nullable" fields must be explicitly set to be considered populated:
//   - singular message fields (whose fields may be unpopulated/default values)
//   - member fields of a oneof (may be their default value)
//   - proto3 optional fields (may be their default value)
//   - proto2 scalar fields (both optional and required)
// - proto3 scalar fields must be non-zero to be considered populated
// - repeated and map fields must be non-empty to be considered populated
//
// ```proto
// message MyMessage {
//   // The field `value` must be set to a non-null value.
//   optional MyOtherMessage value = 1 [(buf.validate.field).required = true];
// }
// ```
```

backend/protos/mgmt/v1alpha1/anonymization.proto
```diff
@@ -27,8 +33,15 @@ message AnonymizeManyResponse {
...
+}
+
 message DefaultTransformersConfig {
-  // Boolean
+  // BooleanP
```
nickzelei, line 44: Guessing you didn't mean to add the `P` to the end of this.

nickzelei (changes requested):

internal/json-anonymizer/json-anonymizer.go
```diff
@@ -147,18 +147,18 @@ func (a *JsonAnonymizer) buildJqQuery() (query string, transformerFunctions []st
 	queryParts := []string{}
 	functionNames := []string{}
 
+	if a.defaultTransformers != nil {
```
nickzelei, line 150: moved this out of the for loop. Guessing this reduced the amount of query parts added? Did this solve a bug or anything?

nickzelei (approved):
overall, lgtm. left one last comment.

internal/json-anonymizer/json-anonymizer.go
alishakawaguchi: Apply defaults was being run last and that causes any mappings in the transformer mappings that are using an expression to be overwritten (paths are okay because we know to skip them). If apply default is run first then expressions applied second it allows users to use more specific generators with expressions. This came up because I was writing the docs

ex
before change:
1. apply expression transformer generate first name to all name fields
2. apply default transformers for string character scramble
this case user ends up with only character scramble fields

after
1. apply default transformers for string character scramble
2. apply expression transformer generate first name to all name fields
users ends up with name fields that look like real names, with the caveat that they have been transformed twice

nickzelei: Won't that cause any configs that are expecting to run Transformers (instead of Generators) on input data to transform generated data instead of the original data?

alishakawaguchi: yes that's correct but only when using JQ expressions. The real fix is figuring out all the paths hit by expressions

nickzelei: Sounds good

alishakawaguchi: made a ticket
https://linear.app/nucleus-cloud/issue/NEOS-1521/json-anonymizer-jq-expressions-find-all-paths-and-add-them-to-skip

PR #2784, opened 2024-10-09
NEOS-1513: Update upgrade and warning prompts

evisdrenova:
This PR:
- deprecates current `reason` field in the isAccountStatusValidResponse in favor of a new typed field called invalid_reason which has an enum of errors. This allows us to use a conditional in the front end to show different UIs based on the error instead of having to parse a string inside of the `Upgrade` component (and anywhere else)
- creates a new `AccountStatusHandler` component which is a wrapper around the `Upgrade` component and `RecordsProgressBar` component. This handles fetching the metrics data it needs and passes into both components while checking to see if it's a neosync cloud account or local etc. 
- Combines the previously separate warning button and upgrade button into one component that conditionally renders the upgrade button. 
- Adds some styling to the errors inside of the upgrade Dialog

## Demo
(the description will be the description from the backend)

https://www.loom.com/share/416c69fe2e0747be830d52db3150b89c?sid=566b9264-1e05-4858-82ad-3e11381f615d


backend/protos/mgmt/v1alpha1/user_account.proto
```diff
@@ -226,12 +226,31 @@ message IsAccountStatusValidRequest {
...
   optional string reason = 2;
   // Whether or not the process should decide to continue polling for validitiy updates
   bool should_poll = 3;
+  // A count of the currently used records for the current billing period.
+  // This may go over the allowed record count depending on when the record count is polled by the metric system.
+  optional uint64 used_record_count = 4;
+  // The allowed record count. It will be null if there is no limit.
+  optional uint64 allowed_record_count = 5;
+  // The current status of the account. Default is valid.
+  optional AccountStatus account_status = 6;
```
nickzelei, line 242: I think `used_record_count` and `account_status` can probably get away with not being optional since we have sane defaults for them.

backend/services/mgmt/v1alpha1/integration_tests/user-account-service_integration_test.go
```diff
@@ -525,7 +525,7 @@ func (s *IntegrationTestSuite) Test_UserAccountService_IsAccountStatusValid_Neos
 	requireNoErrResp(s.T(), resp, err)
 
 	require.False(s.T(), resp.Msg.GetIsValid())
-	require.NotEmpty(s.T(), resp.Msg.GetReason())
+	require.NotEmpty(s.T(), resp.Msg.GetAccountStatus())
```
nickzelei, line 528: Does this need to change?
If anything I would expect this to be unchanged, but new lines added for the expected state of the new additions (used, allowed count and the account status)

backend/services/mgmt/v1alpha1/user-account-service/billing.go
```diff
@@ -170,21 +170,27 @@ func (s *Service) IsAccountStatusValid(
...
 			requested := req.Msg.GetRequestedRecordCount()
 			totalUsed := currentUsed + requested
 			if totalUsed > allowed {
-				reason := fmt.Sprintf("Adding requested record count (%d) would exceed the allowed limit (%d). Current used: %d.", requested, allowed, currentUsed)
+				description := fmt.Sprintf("Adding requested record count (%d) would exceed the allowed limit (%d). Current used: %d.", requested, allowed, currentUsed)
 				return connect.NewResponse(&mgmtv1alpha1.IsAccountStatusValidResponse{
-					IsValid: false,
-					Reason:  &reason,
+					IsValid:            false,
+					AccountStatus:      mgmtv1alpha1.AccountStatus_ACCOUNT_STATUS_REQUESTED_EXCEEDS_LIMIT.Enum(),
```
nickzelei, line 190: Hm, I'm not sure we want to call `.Enum()` here, that shouldn't be necessary.

frontend/apps/web/components/site-header/AccountStatusHandler.tsx
```diff
@@ -0,0 +1,52 @@
...
+import RecordsProgressBar from './RecordsProgressBar';
+import Upgrade from './Upgrade';
+
+interface Props {
+  systemAppConfig: SystemAppConfig;
+}
+
+export function AccountStatusHandler(props: Props) {
+  const { account } = useAccount();
+  const { systemAppConfig } = props;
```
nickzelei, line 17: This is a nit but it's best practice to have the props be on the first line.

frontend/apps/web/components/site-header/Upgrade.tsx
```diff
@@ -36,47 +51,69 @@ export default function Upgrade(props: UpgradeProps): ReactElement | null {
...
-        reason={isAccountStatusValidResp?.reason}
-      />
-      <UpgradeButton href={billingHref} />
+      {!isAccountStatusValidResp?.isValid ? (
+        <UpgradeInfoDialog
+          upgradeHref={billingHref}
+          reason={isAccountStatusValidResp?.accountStatus}
+          description={isAccountStatusValidResp?.reason}
+          count={count}
+          allowedRecords={allowedRecords}
```
nickzelei, line 69: A request to the `isAccountStatusValid` endpoint is being made in this component, so the values can be pulled from here.

OTOH, you could just pass the result down from the parent component since it seems it is being called up there and passing down some of its data to this component.

If you did that, this component would become less smart, which is nice :)

frontend/apps/web/components/site-header/Upgrade.tsx
```diff
@@ -36,47 +51,69 @@ export default function Upgrade(props: UpgradeProps): ReactElement | null {
...
     </div>
   );
 }
 
 interface UpgradeInfoDialogProps {
   upgradeHref: string;
-  reason?: string;
+  count: number;
+  allowedRecords: number;
+  reason?: AccountStatus;
```
nickzelei, line 82:
```suggestion
  accountStatus?: AccountStatus;
```

frontend/apps/web/components/site-header/Upgrade.tsx
```diff
@@ -94,15 +131,95 @@ function UpgradeInfoDialog(props: UpgradeInfoDialogProps): ReactElement {
 }
 
 interface IncludedReasonProps {
-  reason: string;
+  reason: AccountStatus;
```
nickzelei, line 134:
```suggestion
  accountStatus: AccountStatus;
```

frontend/apps/web/components/site-header/Upgrade.tsx
```diff
@@ -94,15 +131,95 @@ function UpgradeInfoDialog(props: UpgradeInfoDialogProps): ReactElement {
...
+      );
+    case AccountStatus.ACCOUNT_IN_EXPIRED_STATE:
+      return (
+        <Alert>
+          <IoAlertCircleOutline className="h-4 w-4" />
+          <AlertTitle>Account is Expired</AlertTitle>
+          <AlertDescription>{description}</AlertDescription>
+        </Alert>
+      );
+    case AccountStatus.REASON_UNSPECIFIED:
```
nickzelei, line 170: I think we updated the thinking that if this is unspecified, the account is actually in a valid state.

nickzelei (changes requested):
Looking good, left a few cleanup comments.

backend/services/mgmt/v1alpha1/integration_tests/user-account-service_integration_test.go
evisdrenova: updated tests

backend/services/mgmt/v1alpha1/user-account-service/billing.go
evisdrenova: it was only necessary since it was optional but now that it;s required we can just return back the value

frontend/apps/web/components/site-header/Upgrade.tsx
evisdrenova: woops totally missed that - yes definitely pass from parent!

backend/services/mgmt/v1alpha1/integration_tests/user-account-service_integration_test.go
```diff
@@ -554,6 +557,9 @@ func (s *IntegrationTestSuite) Test_UserAccountService_IsAccountStatusValid_Neos
 
 		require.False(s.T(), resp.Msg.GetIsValid())
 		require.NotEmpty(s.T(), resp.Msg.GetReason())
+		require.NotEmpty(s.T(), resp.Msg.GetAccountStatus())
+		require.Equal(s.T(), uint64(101), resp.Msg.GetUsedRecordCount())
```
nickzelei, line 561: This should just be 50 since that is what the query is returning. The requested count does not change the used record count.

evisdrenova: beat me to it!

backend/services/mgmt/v1alpha1/integration_tests/user-account-service_integration_test.go
```diff
@@ -526,6 +526,9 @@ func (s *IntegrationTestSuite) Test_UserAccountService_IsAccountStatusValid_Neos
 
 	require.False(s.T(), resp.Msg.GetIsValid())
 	require.NotEmpty(s.T(), resp.Msg.GetReason())
+	require.NotEmpty(s.T(), resp.Msg.GetAccountStatus())
```
nickzelei, line 529: I know you just got these tests passing, but all of these isaccountstatusvalid tests should be asserting what the expected AccountStatus is for each test. Simply asserting that it is not empty isn't really good enough here when we have bespoke tests that are driving a specific account status. NotEmpty is kind of a noop because it's not an optional return type.

nickzelei: We have 9 different `IsAccountStatusValid` tests, each one should have an assertion on what the expected account status is so that we know we got it right.

nickzelei (changes requested):
looks good, one last comment on the account status for integration tests

backend/services/mgmt/v1alpha1/integration_tests/user-account-service_integration_test.go
evisdrenova: good feedback - makes sense - updated

nickzelei (approved):

PR #2792, opened 2024-10-10
NEOS-1517: adds advanced s3 destination options to job configuration

nickzelei:
image

- adds a variety of s3 destination config options
- updates the file names in s3 to be `records-<count>-<timestamp>.jsonl.gz`
- explicitly sets the content type of the record on the bucket to the gz mimetype.


frontend/apps/web/app/(mgmt)/[account]/connections/[id]/components/AwsS3Form.tsx
```diff
@@ -80,223 +80,227 @@ export default function AwsS3Form(props: Props) {
     }
   }
   return (
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/connections/[id]/components/AwsS3Form.tsx:82]: hide whitespace in the PR to get a better sense of what changed here.
I changed some nesting and reduced the spacing that was added as a part of the form error change to squish the whitespace a bit more.

frontend/apps/web/components/jobs/Form/DestinationOptionsForm.tsx
```diff
@@ -18,6 +20,7 @@ interface DestinationOptionsProps {
   hideInitTableSchema?: boolean;
   hideDynamoDbTableMappings?: boolean;
   destinationDetailsRecord: Record<string, DestinationDetails>;
+  errors?: FieldErrors<DestinationOptionsFormValues>;
```
nickzelei, line 23: Only sub component that is using this right now is AwsS3, but this sets us up nicely to update the other destination forms to not directly depend on react-hook-form.

alishakawaguchi (approved):

PR #2795, opened 2024-10-11
NEOS-1516: updates to not show any connect errors for unsupported connections

nickzelei:
This PR does two things:

- Updates the UI to no longer show the Connection error for any unsupported connections
- Fixes issue in the test connection badge where we were double nested <a /> tags

image

What is shown today in prod for S3 destinations:
image

Now nothing is shown:
image

Here is an image of the failed connect badge after removing the double anchor (to show there are no style changes):
image

### DIFF
--- a/frontend/apps/web/components/connections/TestConnectionBadge.tsx
+++ b/frontend/apps/web/components/connections/TestConnectionBadge.tsx
@@ -7,18 +7,18 @@ import { TiWarningOutline } from 'react-icons/ti';
 
 interface TestConnectionBadgeProps {
   validationResponse: CheckConnectionConfigResponse | undefined;
-  id: string | undefined;
+  connectionId: string | undefined;
   accountName: string;
 }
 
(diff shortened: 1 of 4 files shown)


frontend/apps/web/components/connections/TestConnectionBadge.tsx
```diff
@@ -2,93 +2,90 @@
...
   validationResponse: CheckConnectionConfigResponse | undefined;
   accountName: string;
-  id: string;
+  connectionId: string;
 }
 
-function ValidationResponseBadge(props: ValidationResponseBadgeProps) {
-  const { validationResponse, accountName, id } = props;
-  const url = `/${accountName}/connections/${id}/permissions`;
+function ValidationResponseBadge(
```
nickzelei, line 35: reshuffled the logic here to more clearly represent code paths. Also added check to just return null if there is no validation response

evisdrenova (approved):

PR #2799, opened 2024-10-12
NOES-1523:adding an expanded modal to the runs timeline

evisdrenova:
Adds an expandable modal to each row in the Run Timeline table. When expanded, there are additional details about the subtasks of the activity, when they run and how long they took. 

The start time has been updated to actually be the scheduled time now instead of the original start time. This is so the duration of the activity aligns with the individual sub tasks when we render them. 

image

### DIFF
--- a/frontend/apps/web/components/RunTImeline/RunTimeline.tsx
+++ b/frontend/apps/web/components/RunTImeline/RunTimeline.tsx
@@ -662,13 +662,6 @@ function ExpandedRowBody(props: ExpandedRowBodyProps): ReactElement {
     }
   };
 
-  const formatTimeDifference = (diff: number): string => {
-    if (diff < 1000) return `+${diff}ms`;
-    if (diff < 60000) return `+${Math.floor(diff / 1000)}s`;
-    if (diff < 3600000) return `+${Math.floor(diff / 60000)}m`;
-    return `+${Math.floor(diff / 3600000)}h`;
-  };
-
   return (
     <div className="flex flex-col w-full h-[124px] p-2  text-sm border-t border-gray-200 dark:border-gray-700 gap-2">
       {task.tasks.map((subtask, index) => (
--- a/frontend/apps/web/components/RunTImeline/RunTimeline.tsx
+++ b/frontend/apps/web/components/RunTImeline/RunTimeline.tsx
@@ -654,24 +654,36 @@ function ExpandedRowBody(props: ExpandedRowBodyProps): ReactElement {
     }
   };
 
+  const formatTimeDifference = (diff: number): string => {
+    if (diff < 1000) return `+${diff}ms`;
+    if (diff < 60000) return `+${Math.floor(diff / 1000)}s`;
+    if (diff < 3600000) return `+${Math.floor(diff / 60000)}m`;
+    return `+${Math.floor(diff / 3600000)}h`;
+  };
+
   return (
-    <div className="flex flex-col h-[124px] p-2 border-t border-gray-200 dark:border-gray-700 items-center gap-2">
-      <div className="flex flex-col w-full py-2 text-sm">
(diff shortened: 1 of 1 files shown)


frontend/apps/web/components/RunTImeline/RunTimeline.tsx
```diff
@@ -56,9 +61,12 @@ export default function RunTimeline(props: Props): ReactElement {
     let endTime = -Infinity;
 
     tasks.forEach((t) => {
+      const scheduled = t.tasks.find(
+        (st) => st.type == 'ActivityTaskScheduled'
+      )?.eventTime;
       startTime = Math.min(
         startTime,
-        convertTimestampToDate(t.startTime).getTime()
+        convertTimestampToDate(scheduled).getTime()
```
nickzelei, line 69: Hm, can you explain why there is a specific edge case here for the scheduled task?

frontend/apps/web/components/RunTImeline/RunTimeline.tsx
```diff
@@ -151,139 +147,249 @@ export default function RunTimeline(props: Props): ReactElement {
...
+    totalDuration,
+    topOffset,
+    toggleExpandedRowBody,
+    isExpanded,
+    isLastItem,
+  } = props;
+
+  const scheduled = task.tasks.find(
+    (st) => st.type == 'ActivityTaskScheduled'
+  )?.eventTime;
```
nickzelei, line 261: same question again here. why are we always looking at this as opposed to just looking at the tasks start time?

nickzelei (commented):
left a comment regarding the time ccalc.

Based on the SS I think this looks great!
All of the math makes me squeamish though 😝

nickzelei:
also providing a loom or something of it in action would be great to see the expansion/collapse, etc.

frontend/apps/web/components/RunTImeline/RunTimeline.tsx
evisdrenova: there isn't so much of an edge case but i did it to match up with the metadata in the screenshot.

for ex. this is how it was before

image

and this is it after the change:

image

in the before, the  scheduled -> start -> completed doesn't match up with the label on the timeline bar above "50" but that doesn't account for the time from when it was scheduled to when it started. so it looks like a little weird since in the metadata details its technically 53 seconds including the time to schedule

in the after, the scheduled -> start -> completed follows the label on the timeline bar above "53" ms since that label is calculating the duration from scheduled -> close. 

we could also just update the timeline label and leave the metadata as in the before it was more of a judgment call where i felt like in the update it accounted for the time from scheduled -> start and in the before it didn't and i felt like it was more inclusive. 

open to suggestions
scheduled: 
start:

frontend/apps/web/components/RunTImeline/RunTimeline.tsx
evisdrenova: explanation above ^^

frontend/apps/web/components/RunTImeline/RunTimeline.tsx
```diff
@@ -465,6 +571,8 @@ function getTaskStatus(
       case 'ActivityTaskCancelRequested':
         isCanceled = true;
         break;
+      case 'ActivityTaskScheduled':
```
TimelyToga, line 574: this switch statement does not include `ActivityTaskStarted` which is mentioned on `L652`

evisdrenova: originally i was going to say it shouldn't need it because we only want to return a separate status from `running` if its complete or failed or canceled or timedout, but for posterity i'll add it to have coverage

nickzelei (approved):
🚢

PR #2808, opened 2024-10-15
NEOS-1545: Update TS API Docs

evisdrenova:
Updated typescript API docs with a little bit of a restructure and a few more examples.

### DIFF
--- a/docs/protos/go.md
+++ b/docs/protos/go.md
@@ -8,40 +8,38 @@ slug: /go
 
 ## Introduction
 
-The Neosync Go SDK is publicly available and can be added to any Go-based project by running `go get github.com/nucleuscloud/neosync`.
+The Neosync Go SDK is publicly available and can be added to any Go project. With the Neosync Go SDK, you can:
 
-Neosync's CLI is the primary user of the Go SDK today, and can be referenced for examples of how to use the SDK.
+1. Anonymize structured data and generate synthetic data
+2. Anonymize free-form text data
+3. Create resources in Neosync such as Jobs, Connections, Transformers and more
 
-## Configuration
+## Installation
 
-There are a few inputs that the SDK needs in order to be properly configured.
+You can add the Neosync Go SDK using:
 
-1. API URL
-2. Account ID
-3. API Key (required for Neosync Cloud or self-hosted authenticated environments)
+`go get github.com/nucleuscloud/neosync`.
 
-### API Url
+## Prerequisites
 
-If using Neosync Cloud, the backend api url is: `https://neosync-api.svcs.neosync.dev`
+There are a few prerequisites that the SDK needs in order to be properly configured.
 
-The standard localhost url is: `http://localhost:8080`
+| **Properties** | **Details**                                                                                                                                                                 |
+| -------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **API URL**    | Production: `https://neosync-api.svcs.neosync.dev`<br /> Local: `http://localhost:8080`                                                                                     |
+| **Account ID** | The account ID may be necessary for some requests and can be found by going into the `/:accountName/settings` page in the Neosync App                                       |
+| **API Key**    | An access token (API key, or user JWT) must be used to access authenticated Neosync environments. For an API Key, this can be created at `/:accountName/settings/api-keys`. |
 
-### Account ID
+## Authentication
 
-The account ID is necessary for some requests that do not have an obvious identifier like retrieving a list of jobs, or a list of connections.
-This can be found by going into the app on the `/:accountName/settings` page and found in the header.
+If you are using Neosync locally and in unauthenticated mode then there is no authentication required and you can move onto the [Getting Started](go#getting-started) section.
 
-### API Key
+If you are using Neosync locally in `auth mode` or using Neosync Cloud, you can authenticate with the Neosync server using an API URL and API Key. There are two ways to provide the authentication header.
 
-An access token (api key, or user jwt) must be used to access authenticated Neosync environments.
-For an API Key, this can be created at `/:accountName/settings/api-keys`.
+1. Attaching to the HTTP client
+2. Providing an interceptor to the SDK Clients that patch in the header on every request.
 
-## Getting Started
-
-Neosync is made up of a number of different services that live inside of the same process.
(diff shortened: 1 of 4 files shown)


docs/protos/home.md
```diff
@@ -8,10 +8,9 @@ slug: /
 
 ## Introduction
 
-Neosync, like many web services these days, is a full-fledged API. Check out the [architecture](/platform) page for an overview of how the Neosync API fits into the picture.
+Neosync has first-class APIs and SDKs that developers can use to integrate Neosync into their workflow. To learn more about how the Neosync API fits into the overall architecture, check out the Check out the [platform page](/platform).
 
-All of our APIs are generated from Protobuf files. For this reason, a client can easily be generated in any of the many supported languages.
-Internally, we make great use of the Go SDK as well as the TypeScript SDK.
+Neosync API serves up [Connect](https://github.com/connectrpc), which can listen using Connect, gRPC, or HTTP protocols. All of our APIs are generated from Protobuf files and our SDKs call gRPC endpoints. Alternatively, you can use our OpenAPI spec to generate REST clients as well.
```
nickzelei, line 13: `All of our APIs are generated from Protobuf files and our SDKs call gRPC endpoints.`

Maybe a slight wording change here but they don't call the gRPC endpoints. They call the Connect endpoints by default. Each SDK can be configured to specifically call the grpc or REST endpoints however.

Moreover, not sure it's really useful info to say that our SDKs call the API endpoints. Is that not implied? (maybe it isn't but if they don't do that, then what are they doing?)

nickzelei: > Alternatively, you can use our OpenAPI spec to generate REST clients as well.

We don't really have this yet (soon), so is it worth leaving out until we do?

docs/protos/typescript.md
```diff
@@ -33,81 +36,280 @@ There are a few inputs that the SDK needs in order to be properly configured.
...
 
-An access token (api key, or user jwt) must be used to access authenticated Neosync environments.
+An access token (API key, or user JWT) must be used to access authenticated Neosync environments.
 For an API Key, this can be created at `/:accountName/settings/api-keys`.
 
-## Usage
-
 For a prime example of how to us this SDK, view the [withNeosyncContext](https://github.com/nucleuscloud/neosync/blob/main/frontend/apps/web/api-only/neosync-context.ts#L23) method in the Neosync app's BFF layer.
 
-### Note on Transports
```
nickzelei, line 54: This is actually a really important section and will vary based on the environment that the user is invoking our SDK in, which is why it has its own section here and imo should not be removed.

The examples in these docs assume node (the examples show importing from the connect-node lib`, but for example, our frontend uses the connect-web since it's called from the client.

docs/protos/typescript.md
```diff
@@ -33,81 +36,280 @@ There are a few inputs that the SDK needs in order to be properly configured.
...
+        transformerMappings: transformers, // pass in your transformer mappings that you defined
+      });
+    console.log('Anonymization result:', result.outputData);
+  } catch (error) {
+    console.error('Error:', error);
+  }
+}
+
+// calling our async function
+runAnonymization();
```
nickzelei, line 187: It would be helpful to update this line to actually await the promise as otherwise the script will exit immeditaely and not wait for the response.
Maybe I'm wrong though.

Something like:
```suggestion
runAnonymization().then(() => console.log("script completed").catch(console.error);
```

docs/protos/typescript.md
```diff
@@ -33,81 +36,280 @@ There are a few inputs that the SDK needs in order to be properly configured.
...
+    });
+    // returning the job.id here
+    console.log('Job result:', result.job?.id);
+  } catch (error) {
+    console.error('Error:', error);
+  }
+}
+
+// calling our async function
+createJob();
```
nickzelei, line 308: Same comment here. But maybe you ran this and didn't see any issue with it.

nickzelei (changes requested):
Nice updates! Left a few comments.

docs/protos/typescript.md
evisdrenova: i guess there could be an error that isn't caught by the internal try-catch so you're prob right that I should just handle that outside of the function - but fwiw i think the presence of the await inside of the try catch will wait for the response - will update anyways

docs/protos/typescript.md
evisdrenova: ah i see good catch

docs/protos/go.md
```diff
@@ -59,22 +56,49 @@ import (
...
+	for key, value := range t.Headers {
+		req.Header.Add(key, value)
+	}
+	return t.Transport.RoundTrip(req)
 }
+
 ```
 
-## Go SDK Packages
+## Getting started
```
nickzelei, line 97: small thing but I wonder if this should be above the authentication section?

docs/protos/home.md
```diff
@@ -10,7 +10,7 @@ slug: /
 
 Neosync has first-class APIs and SDKs that developers can use to integrate Neosync into their workflow. To learn more about how the Neosync API fits into the overall architecture, check out the Check out the [platform page](/platform).
 
-Neosync API serves up [Connect](https://github.com/connectrpc), which can listen using Connect, gRPC, or HTTP protocols. All of our APIs are generated from Protobuf files and our SDKs call gRPC endpoints. Alternatively, you can use our OpenAPI spec to generate REST clients as well.
+Neosync API serves up [Connect](https://github.com/connectrpc), which can listen using Connect, gRPC, or HTTP protocols. All of our APIs are generated from Protobuf files and our SDKs call Connect endpoints by default. Each SDK can be configured to call the gRPC or REST endpoints in Connect.
```
nickzelei, line 13: Maybe: `Each SDK can be configured to use gRPC or REST in lieu of the default (Connect).`?

nickzelei (approved):
left two non blocking comments.

docs/protos/go.md
evisdrenova: traditionally authentication comes first, since you install -> authenticate -> then do something

nickzelei: yea it's just a little weird because it shows a full example, and then it goes into the "getting started" section that talks all about the makeup of the packages.
I think that was why I previously had the auth in its own section separate.

nickzelei (approved):

PR #2811, opened 2024-10-15
Add debug flag to cli sync

alishakawaguchi:
this is just a start. need to add more debug statements


nickzelei (approved):
LGTM, over time we should maybe update the CLI everywhere to just use the charm logger if it's a good one?
The Debug flag might be good to have as a global parameter too set at the root.

Alternatively, we could also provide the ability to set a log level instead?

I do like the `--debug` flag though. Really clear.

PR #2817, opened 2024-10-16
Makes all transformer arugments optional and adds tests

alishakawaguchi:
makes transformer config parameters optional in proto
Fixes NEOS-1543
Fixes NEOS-1537
Fixes NEOS-1536
Fixes NEOS-1535
Fixes NEOS-1533
Fixes NEOS-1534


worker/pkg/workflows/datasync/activities/gen-benthos-configs/processors.go
```diff
@@ -435,7 +435,7 @@ func computeMutationFunction(col *mgmtv1alpha1.JobMapping, colInfo *sqlmanager_s
 
 	switch col.Transformer.Source {
```
nickzelei, line 436: Main question I have here:

If these now all have presence detection, using the Getters removes that check right, so it will always be the system default?

Seems ok for bools, empty arrays, etc, but for min/max it is kinda weird, no?

```
minValue := col.Transformer.Config.GetGenerateFloat64Config().GetMin()
maxValue := col.Transformer.Config.GetGenerateFloat64Config().GetMax()
```

If none of these are provided, 0, 0 will be passed in to the transformer.
Is that desired, or would we now need to conditionally build the blobl method to handle omitting those values so that it falls back to the transformer defined default value?


```
generateFullName := col.Transformer.Config.GetGenerateCountryConfig().GetGenerateFullName()
return fmt.Sprintf(`generate_country(generate_full_name:%t)`, generateFullName), nil
```
Would result in an empty generateFullName being passed in.

nickzelei: If you wanna chat more tomorrow after standup.

backend/services/mgmt/v1alpha1/integration_tests/anonymization-service_integration_test.go
```diff
@@ -85,58 +85,6 @@ func (s *IntegrationTestSuite) Test_AnonymizeService_AnonymizeMany() {
   "active": false,
   "sports": ["basketball", "golf", "swimming"]
 }`,
-	}
-
-	resp, err := s.unauthdClients.anonymize.AnonymizeMany(
```
nickzelei: just curious, but why was this call removed?

worker/pkg/benthos/transformers/generate_random_string.go
```diff
@@ -55,17 +55,28 @@ func init() {
...
 		return NewGenerateRandomStringOpts(
 			nil,
-			nil,
+			maxLen,
 			nil,
 		)
 	}
+	minValue := config.Min
+	maxValue := config.Max
+	if maxLen != nil && maxValue != nil {
```
nickzelei, line 68: Why is the maxValue only set if both maxLen and maxValue are not nil?

worker/pkg/benthos/transformers/generator_utils.go
```diff
@@ -139,24 +143,44 @@ func ParseBloblangSpec(benthosSpec *BenthosSpec) (*ParsedBenthosSpec, error) {
...
+	// Looks for bloblang.RegisterFunctionV2 and captures the function name in quotes
+	re := regexp.MustCompile(`bloblang\.RegisterFunctionV2\("([^"]+)"`)
+
+	matches := re.FindStringSubmatch(input)
+
+	if len(matches) > 1 {
+		return matches[1], nil
+	}
+
+	return "", fmt.Errorf("bloblang function name not found")
```
nickzelei, line 181: Is there a way this can say what the function name is? Is that the input? just curious if it would make this easier to understand ifwe knew what functionname was failed to be found.

nickzelei (approved):
overall this LGTM, left a few straggling comments but I don't think they are blocking. More informational.

backend/services/mgmt/v1alpha1/integration_tests/anonymization-service_integration_test.go
alishakawaguchi: hmm I don't see this in my changes

PR #2818, opened 2024-10-16
AWS S3 getconnectionschema handle empty table folders

alishakawaguchi:


backend/services/mgmt/v1alpha1/connection-data-service/connection-data.go
```diff
@@ -627,7 +627,7 @@ func (s *Service) GetConnectionSchema(
 					return nil, err
 				}
 				if out == nil {
-					break
+					continue
```
nickzelei, line 630: maybe we should add a nice log statement here ? It might be nice if it's an edge case.

backend/services/mgmt/v1alpha1/connection-data-service/connection-data.go
```diff
@@ -637,6 +637,9 @@ func (s *Service) GetConnectionSchema(
 				if err != nil {
 					return nil, err
 				}
+				if result.ContentLength == nil || *result.ContentLength == 0 {
+					continue
```
nickzelei, line 641: Same, is it worth adding a log statement here? It might be helpful for debugging if this is an edge case.

nickzelei (changes requested):

nickzelei (approved):

PR #2836, opened 2024-10-22
NEOS-1560: expose entities in api for transformpiitext

nickzelei:
- adds new rpc in transformers service that exposes available pii entities
- updates transform_pii_text config to allow specifying allowed entities


backend/protos/mgmt/v1alpha1/transformer.proto
```diff
@@ -501,4 +519,9 @@ service TransformersService {
   rpc IsTransformerNameAvailable(IsTransformerNameAvailableRequest) returns (IsTransformerNameAvailableResponse) {}
   rpc ValidateUserJavascriptCode(ValidateUserJavascriptCodeRequest) returns (ValidateUserJavascriptCodeResponse) {}
   rpc ValidateUserRegexCode(ValidateUserRegexCodeRequest) returns (ValidateUserRegexCodeResponse) {}
+
+  // Retrieve a list of available Pii recognizers for use with the TransformPiiText transformer
+  rpc GetTransformPiiRecognizers(GetTransformPiiRecognizersRequest) returns (GetTransformPiiRecognizersResponse) {
```
nickzelei, line 524: I put this here because it seemed more generic than just shoving it in the anonymization service since it is specific to the transformPiitext transformer.
wdyt @alishakawaguchi, @evisdrenova  ?

backend/protos/mgmt/v1alpha1/transformer.proto
```diff
@@ -232,6 +232,15 @@ message TransformPiiText {
 
   // Configure deny lists where each word is treated as PII.
   repeated PiiDenyRecognizer deny_recognizers = 3;
+
+  // Configure a list of recognizers to be used for PII analysis. If not provided or empty, all recognizers are used
+  // If this is specified, any ad-hoc, or deny_recognizers names must also be provided.
+  // To see available builtin recognizers, call the GetPiiTextRecognizers() RPC method to see what is available for your account.
+  repeated string allowed_recognizers = 4 [(buf.validate.field).repeated = {
```
nickzelei, line 239: What do we think about this name?

Will it be confusing when looking at the property above called `deny_recognizers`?

Should we have a different name here?
wdyt: @alishakawaguchi @evisdrenova  ?

Maybe it makes more sense once we rename the `deny_recognizers` to just be `recognizers` or `adhoc_recognizers`

backend/protos/mgmt/v1alpha1/transformer.proto
evisdrenova: Is this referring to the getEntities endpoint that we talked about or something else? If so, are we just using recognizers instead of entities everywhere? I like entities because that's industry standard i.e. Named Entity Recognition model.

nickzelei: oh yeah actually good call, I'm surfacing the wrong function. I should be surfacing the entities, not the entity recognziers.

backend/protos/mgmt/v1alpha1/transformer.proto
evisdrenova: Hmm I think it shoudl be called something different since it's not technically a recognizer as we have it defined in the `deny_recognizer` from a type perspective. I think it might be be clearer to call it something like `allowed_terms` or `allowed_words` or something along that line?

backend/protos/mgmt/v1alpha1/transformer.proto
nickzelei: That is kind of beside my original question though.

backend/protos/mgmt/v1alpha1/transformer.proto
nickzelei: sorry I was calling it the wrong thing as I was exposing the wrong function. this is will more clearly be named `allowed_entities` now

backend/protos/mgmt/v1alpha1/transformer.proto
evisdrenova: yeah back to your original point - IMO i'd put it in the anonymizer service since it retrieves the entities from the presidio api and functionally is currently only used for anonymization even if it is consumed by a transformer

nickzelei: only problem is that ceases to be true once we wire it up to the async job flow. it's a supporting value for a transformer

evisdrenova: which i think is okay imo - you're still getting those values from presidio it's just being used by the transformer, for ex. we have the connection rpcs in the connection_data service even though it's an input into a job. but i think you can go either way with it

backend/services/mgmt/v1alpha1/transformers-service/entities.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := s.entityclient.GetSupportedentitiesWithResponse(ctx, &presidioapi.GetSupportedentitiesParams{
+		Language: &enLanguage,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("was unable to retrieve available recognizers: %w", err)
+	}
```
evisdrenova, line 39:
```suggestion
		return nil, fmt.Errorf("was unable to retrieve available entities: %w", err)
	}
```

backend/services/mgmt/v1alpha1/transformers-service/entities.go
```diff
@@ -0,0 +1,48 @@
...
+	if resp.JSON200 == nil {
+		return nil, fmt.Errorf("received non-200 response from recognizer api: %s %d %s", resp.Status(), resp.StatusCode(), string(resp.Body))
+	}
```
evisdrenova, line 42:
```suggestion
		return nil, fmt.Errorf("received non-200 response from entities api: %s %d %s", resp.Status(), resp.StatusCode(), string(resp.Body))
	}
```

backend/services/mgmt/v1alpha1/transformers-service/system_transformers.go
```diff
@@ -628,38 +628,61 @@ var (
 		},
 	}
 
-	systemTransformerSourceMap = map[mgmtv1alpha1.TransformerSource]*mgmtv1alpha1.SystemTransformer{}
+	// base transformers + ee transformers
+	allSystemTransformers = []*mgmtv1alpha1.SystemTransformer{}
+
+	baseSystemTransformerSourceMap = map[mgmtv1alpha1.TransformerSource]*mgmtv1alpha1.SystemTransformer{}
+
+	allSystemTransformersSourceMap = map[mgmtv1alpha1.TransformerSource]*mgmtv1alpha1.SystemTransformer{}
```
evisdrenova, line 636: later would be nice to show all the transformers but have some disabled with like an "upgrade now" button to incentivize users to upgrade in order to get access to those transformers.

evisdrenova (approved):
LGTM - just left a couple of small comments

backend/protos/mgmt/v1alpha1/transformer.proto
alishakawaguchi: I think the transformer service makes the most sense

alishakawaguchi (approved):

PR #2838, opened 2024-10-23
NEOS-1540: add support for allow list in anonymizer

nickzelei:


backend/protos/mgmt/v1alpha1/transformer.proto
```diff
@@ -241,6 +241,9 @@ message TransformPiiText {
       string: {min_len: 1}
     }
   }];
+
+  // Exact match of PII phrases identified by the analysis engine.
+  repeated string allowed_phrases = 5;
```
nickzelei, line 246: Can we think of a better name here? I liked `allowed_phrases` better than `allow_words` because I feel that it illustrates the point better. But maybe we could do even better? wdyt @evisdrenova

alishakawaguchi (approved):

PR #2840, opened 2024-10-23
Use json decoder instead of scanner when reading from document storage

alishakawaguchi:


nickzelei (commented):
talked offline, one bug with the error decoder check

nickzelei (approved):

PR #2871, opened 2024-10-28
NEOS-1579: add free trial timer to header

evisdrenova:
This feature adds a free trial countdown to the top nav menu to give users an idea of how many days they have left in their trial. 

image


backend/protos/mgmt/v1alpha1/user_account.proto
```diff
@@ -246,6 +246,8 @@ message IsAccountStatusValidResponse {
   optional uint64 allowed_record_count = 5;
   // The current status of the account. Default is valid.
   AccountStatus account_status = 6;
+  // The timestamp of when the account
```
nickzelei, line 249:
```suggestion
  // The timestamp of when the account was created.
```

backend/services/mgmt/v1alpha1/user-account-service/billing.go
```diff
@@ -150,6 +151,11 @@ func (s *Service) IsAccountStatusValid(
 	ctx context.Context,
 	req *connect.Request[mgmtv1alpha1.IsAccountStatusValidRequest],
 ) (*connect.Response[mgmtv1alpha1.IsAccountStatusValidResponse], error) {
+	accountId, err := s.verifyUserInAccount(ctx, req.Msg.GetAccountId())
+	if err != nil {
+		return nil, err
+	}
```
nickzelei, line 157: This isn't really necessary here because the `GetAccountStatus` handles it.

frontend/apps/web/components/site-header/AccountStatusHandler.tsx
```diff
@@ -35,3 +45,69 @@ export function AccountStatusHandler(props: Props) {
...
 }
+
+interface TrialCountdownProps {
+  createdDate: number;
+  isAccountStatusValidResp: IsAccountStatusValidResponse | undefined;
+}
+
+const TRIAL_DURATION_DAYS = 14;
+
+// TODO: created_at is coming back as blank
```
nickzelei, line 56: Is this still valid?

frontend/apps/web/components/site-header/AccountStatusHandler.tsx
```diff
@@ -35,3 +45,69 @@ export function AccountStatusHandler(props: Props) {
...
+
+function TrialCountdown(props: TrialCountdownProps) {
+  const { createdDate, isAccountStatusValidResp } = props;
```
nickzelei, line 59: As a heads up, we don't currently store our created at with a timestamp, so it comes back on the server as the date in UTC (which is fine because we handle all of the billing in UTC).
but you'll want to make sure all of the date handling here is done in UTC format.

frontend/apps/web/components/site-header/AccountStatusHandler.tsx
```diff
@@ -35,3 +45,69 @@ export function AccountStatusHandler(props: Props) {
     </div>
   );
 }
+
+interface TrialCountdownProps {
+  createdDate: number;
+  isAccountStatusValidResp: IsAccountStatusValidResponse | undefined;
+}
+
+const TRIAL_DURATION_DAYS = 14;
```
nickzelei, line 54: Ideally we could do most or all of this on the server so we don't have this copied in two places.

frontend/apps/web/components/site-header/AccountStatusHandler.tsx
```diff
@@ -26,6 +28,14 @@ export function AccountStatusHandler(props: Props) {
 
   return (
     <div className="flex flex-row items-center gap-2">
+      {!systemAppConfig.isNeosyncCloud && (
+        <TrialCountdown
```
nickzelei, line 32: Seems like this will show for all accounts regardless of their account type or status? Based on what I see in the component the default state is "expired", so even if you are a valid account it will show this but in expired state? Unless I am reading that wrong.

frontend/apps/web/components/site-header/AccountStatusHandler.tsx
```diff
@@ -35,3 +45,69 @@ export function AccountStatusHandler(props: Props) {
...
+  const { createdDate, isAccountStatusValidResp } = props;
+
+  const now = Date.now();
+  const trialEndDate = new Date(
+    createdDate + TRIAL_DURATION_DAYS * 24 * 60 * 60 * 1000
+  );
+  const daysRemaining = Math.max(
+    0,
+    Math.ceil((trialEndDate.getTime() - now) / (1000 * 60 * 60 * 24))
+  );
```
nickzelei, line 68: Would be cleaner to use date-fns to do most of this for you so that it properly handles time changes like daylight savings, etc.

An example of something more functional:

```
import { differenceInDays, addDays, isAfter } from 'date-fns';


interface TrialStatus {
  isTrialExpired: boolean;
  isNearingEnd: boolean;
  daysRemaining: number;
  trialEndDate: Date;
}

export function checkTrialStatus(createdAt: Date): TrialStatus {
  const TRIAL_DURATION_DAYS = 14;
  const WARNING_THRESHOLD_DAYS = 3;
  
  const currentDate = new Date();
  const trialEndDate = addDays(createdAt, TRIAL_DURATION_DAYS);
  
  const daysRemaining = differenceInDays(trialEndDate, currentDate);
  
  return {
    isTrialExpired: isAfter(currentDate, trialEndDate),
    isNearingEnd: daysRemaining <= WARNING_THRESHOLD_DAYS && daysRemaining > 0,
    daysRemaining: Math.max(0, daysRemaining),
    trialEndDate
  };
}
```

And then to use it:
```ts
const trialStatus = checkTrialStatus(createdAt);

if (trialStatus.isTrialExpired) {
  console.log('Trial has expired');
} else if (trialStatus.isNearingEnd) {
  console.log(`Trial ending soon! ${trialStatus.daysRemaining} days remaining`);
} else {
  console.log(`${trialStatus.daysRemaining} days remaining in trial`);
}
```

nickzelei: I don't think this handles the last day (hours remaining) but you kind of get the point here.

nickzelei (changes requested):
This is pretty cool, left some comments regarding the handling of the logic.

backend/services/mgmt/v1alpha1/user-account-service/billing.go
evisdrenova: updated

frontend/apps/web/components/site-header/AccountStatusHandler.tsx
```diff
@@ -35,3 +50,65 @@ export function AccountStatusHandler(props: Props) {
...
+  const now = Date.now();
+  const daysRemaining = Math.max(
+    0,
+    Math.ceil((trialEndDate - now) / (1000 * 60 * 60 * 24))
+  );
+
+  const isExpired =
+    isAccountStatusValidResp?.accountStatus ==
+    AccountStatus.ACCOUNT_TRIAL_EXPIRED;
+  const isAlmostExpired = daysRemaining <= 3;
```
nickzelei, line 71: I still think this would be clearer if the date-fns were used here as it's going to handle more edge cases.

```
const daysRemaining = differenceInDays(trailEndDate, now);
const isAlmostExpired = daysRemaining <= 3
```

frontend/apps/web/components/site-header/AccountStatusHandler.tsx
```diff
@@ -35,3 +50,65 @@ export function AccountStatusHandler(props: Props) {
     </div>
   );
 }
+
+interface TrialCountdownProps {
+  trialEndDate: number;
+  isAccountStatusValidResp: IsAccountStatusValidResponse | undefined;
```
nickzelei, line 56: might be worth simplifying these props to just pass in an `isExpired` and do the calculation outside of this function. make it less smart.

frontend/apps/web/components/site-header/AccountStatusHandler.tsx
```diff
@@ -35,3 +50,65 @@ export function AccountStatusHandler(props: Props) {
...
+    0,
+    Math.ceil((trialEndDate - now) / (1000 * 60 * 60 * 24))
+  );
+
+  const isExpired =
+    isAccountStatusValidResp?.accountStatus ==
+    AccountStatus.ACCOUNT_TRIAL_EXPIRED;
+  const isAlmostExpired = daysRemaining <= 3;
+
+  if (isExpired)
```
nickzelei, line 73: based on this logic it doesn't look like the `isAlmostExpired` will ever actually be used.
If `isExpired` is true, then `isAlmostExpired` will always be true..

nickzelei (changes requested):
Improvements. Still some logical issues with the handling of isExpired vs isAlmostExpired.
Would also be great to see the date-fns used here instead of manually calculating the difference.

frontend/apps/web/components/site-header/AccountStatusHandler.tsx
evisdrenova: whoops meant to clean that up, that was leftover code

frontend/apps/web/components/site-header/AccountStatusHandler.tsx
```diff
@@ -31,14 +32,19 @@ export function AccountStatusHandler(props: Props) {
     (data?.accountStatus == AccountStatus.ACCOUNT_TRIAL_ACTIVE ||
       data?.accountStatus == AccountStatus.ACCOUNT_TRIAL_EXPIRED);
 
+  const trialEndDate = new Date(
+    data?.trialExpiresAt?.toDate() ?? Date.now()
+  ).getTime();
+
+  const daysRemaining = differenceInDays(Math.max(trialEndDate), Date.now());
```
nickzelei, line 39:
```suggestion
  const daysRemaining = Math.max(0, differenceInDays(trialEndDate, Date.now()));
```

nickzelei (approved):
LGTM, minor thing regarding the math.max

PR #2882, opened 2024-10-30
Updates CLI and Worker to used shared benthos builder

alishakawaguchi:


internal/benthos/benthos-builder/builders/mongodb.go
```diff
@@ -85,26 +87,51 @@ func (b *benthosBuilder) getMongoDbSyncBenthosConfigResponses(
...
+		Operation:  "update-one",
+		Upsert:     true,
+		DocumentMap: `
+						  root = {
+								"$set": this
+							}
+						`,
+		FilterMap: `
+						  root._id = this._id
+						`,
```
nickzelei, line 130: this is rendering so strangely in github. must be the tabs in the string.

nickzelei (approved):

PR #2892, opened 2024-10-31
NEOS-1593: Business name transformer

evisdrenova:
This feature adds in support to randomly generate a realistic looking business name with a new transformer called Generate Business Name. We pre-generate like 1.6M business names as part of the `go generate` script and store them in the `/transformers/data-sets/datasets` directory. 

If we're concerned about that .txt file being too big, we can reduce it but I took the approach that business names are more likely to be required to be unique compared to first or last names, so we should have more than less. But if we want to knock it down to like 100k like the others, happy to update it. 

Demo
https://www.loom.com/share/b7c01ba19b8c433e9bfca8f7b667ab57

Edit:

Did some benchmarking - here's what I'm seeing. 

I checked the build time for each generator in the `generators.go` file as well as mem + cpu usage.

Overall, it doesn't seem to spike memory or cpu, but does take about 4 seconds to run the businessName generator. 
@nickzelei 

| Generator      | Avg Time (s) | Avg Memory (MB) | Avg CPU (%) | Time Range (s) |
|---------------|--------------|-----------------|-------------|----------------|
| emaildomains  | 0.200       | 50.75          | 13.1        | 0.000 - 1.000 |
| first_names   | 0.400       | 50.62          | 12.1        | 0.000 - 1.000 |
| last_names    | 0.600       | 50.59          | 12.8        | 0.000 - 1.000 |
| business_names| 3.800       | 50.83          | 13.9        | 3.000 - 4.000 |


worker/pkg/benthos/transformers/data-sets/data_generator.go
```diff
@@ -68,7 +68,7 @@ func main() {
 
 	for _, line := range lines {
 		trimmedLine := strings.TrimSpace(line)
-		if len(trimmedLine) == 0 || strings.ContainsAny(trimmedLine, " -_") {
+		if len(trimmedLine) == 0 || strings.ContainsAny(trimmedLine, "-_") {
```
nickzelei, line 71: Hm, I'm guessing there was a reason why I originally wrote this to exclude first/last names with spaces. But who knows if that is still relevant. Part of me wants to expose any bugs removing the space here would cause...

nickzelei (approved):
Nice! That is a big add 3 million additions.

Did you find this to affect build times at all?

alishakawaguchi (approved):

PR #2893, opened 2024-11-01
Fix mysql date and datetime in CLI sync

alishakawaguchi:
Fixes
- adds read write mutex to builder registry
- fixes cli sync not exiting properly when hitting an error in sync

Mysql
Inserts but is wrong
- json

Fails to insert
- bit column type
- binary column type


cli/internal/cmds/neosync/sync/sync.go
```diff
@@ -297,7 +297,9 @@ func (c *clisync) configureAndRunSync() error {
 			case <-ctx.Done():
 				return
 			case <-stopChan:
+				c.logger.Error("Sync Failed.")
 				cancel()
+				os.Exit(0)
```
nickzelei, line 302: interesting. I guess because we arent cancelling the outer context. Should probably look into how we can do that so that the entire program can exit.

Also if this is an error should `os.Exit(1)`

worker/pkg/benthos/json/sql_processor.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/warpstreamlabs/bento/public/service"
+)
+
+func sqlToJsonProcessorConfig() *service.ConfigSpec {
+	return service.NewConfigSpec()
+}
+
+func RegisterSqlToJsonProcessor(env *service.Environment) error {
+	return env.RegisterBatchProcessor(
+		"sql_to_json",
```
nickzelei, line 16: Is this mysql specific or sql-generic specifc ? or is the intention to eventually support all 3 within this processor?

backend/pkg/sqlconnect/sql-connector.go
```diff
@@ -34,24 +34,54 @@ type SqlDBTX interface {
...
 
+type SqlConnectorOption func(*sqlConnectorOptions)
+
+type sqlConnectorOptions struct {
+	mysqlDisableParseTime bool
+	postgresDriver        *string
+}
+
+// WithMysqlDisableParseTime disables MySQL time parsing
+func WithMysqlDisableParseTime(disable bool) SqlConnectorOption {
```
nickzelei, line 45: Nit for future, might be clearer if this defualts to true to just have a function like:

```
func WithMysqlParseTimeDisabled() SqlConnectorOption {
  return func(opts *sqlConnectorOptions) {
    opts.mysqlDisableParseTime = true
  }
}
```

backend/pkg/sqlconnect/sql-connector.go
```diff
@@ -34,24 +34,54 @@ type SqlDBTX interface {
...
+	return func(opts *sqlConnectorOptions) {
+		opts.mysqlDisableParseTime = disable
+	}
+}
+
+// WithPostgresDriver overrides default postgres driver
+func WithPostgresDriver(postgresDriver string) SqlConnectorOption {
```
nickzelei, line 52: Similar thing here. Might be clearer for us if we for instance default to pgx, to have an option simply called: `WithDefaultPostgresDriver()` the just does `opts.postgresDriver = "postgres"`

There are few options here so we can make it brain dead simple by just having these functions that fully encapsulate the intent.

nickzelei (approved):
left small comments that I think can be fast follows. The only one that might be worth fixing is the os.exit, but even that is kinda minor.

nickzelei (approved):
clean 🚢

worker/pkg/benthos/json/sql_processor.go
alishakawaguchi: Not entirely sure yet. For now keeping it sql generic until I get a better idea of how much needs to be fixed for postgres and mssql. if the logic get complicated then I plan to split them into driver specific processors

PR #2896, opened 2024-11-01
Update workflow integration tests to use API and run in parallel

alishakawaguchi:


alishakawaguchi:
I can't figure out why the linter is complaining about unused functions when they are all used

nickzelei:
> I can't figure out why the linter is complaining about unused functions when they are all used

it's probably because they live in files that are not `_test.go` but `-test.go` and they are only being used from within `_test.go` files.

That's my hunch anyways. The linter is set to ignore lint errors in _test files (maybe this is a bad move on our part). So it is seeing those functions as being unused.

nickzelei (approved):

nickzelei:
nice
image

PR #2907, opened 2024-11-04
NEOS-1596: add in ip address transformer to backend, worker

evisdrenova:
This features adds in support for an IP Address transformer. It supports:
- IPV6 addresses
- IPV4 addresses - the user can config them to be either private-a, private-b, private-c classes, loopback, linklocal or multicast

## Demo

https://www.loom.com/share/b6eabc19b0c94009a35a2d8fe9456352?sid=bd6fb377-6adc-4508-9835-e533a7c5aa3e


backend/protos/mgmt/v1alpha1/transformer.proto
```diff
@@ -515,6 +517,39 @@ message GetTransformPiiEntitiesResponse {
...
+enum GenerateIpAddressVersion {
+  // Unspecified defaults to ipv4
+  GENERATE_IP_ADDRESS_VERSION_UNSPECIFIED = 0;
+  // Generates a random ipv4
+  GENERATE_IP_ADDRESS_VERSION_V4 = 1;
+  // Generates a random ipv6
+  GENERATE_IP_ADDRESS_VERSION_V6 = 2;
+}
+
+enum GenerateIpAddressClass {
```
nickzelei, line 529: Could we just collapse the `GenerateIpAddressVersion` into this one? I think most of these only apply to Ipv4 anyways, so it doesn't really make sense to be able to configure v6 along with a Class B address. The way this is defined now leads to invalid combinations.

I would suggest either doing a Oneof where you specify either an Ipv6 Config or an Ipv4 Config, or just have a variety of discrete enums for V4 or V6.
Happy to chat more about this offline if it isn't clear.

nickzelei (changes requested):
Let's workshop the proto a little bit, I think that will help further simplify a lot of the code.

worker/pkg/benthos/transformers/generate_ip_address_test.go
```diff
@@ -195,9 +187,9 @@ func Test_IpV4Public_NoOptions(t *testing.T) {
 	assert.False(t, isReservedIP(parsedIP), "should not be reserved IP")
 }
 
-func Test_IpV4PrivateA_Class(t *testing.T) {
-	class := "GENERATE_IP_ADDRESS_CLASS_PRIVATE_A"
-	mapping := fmt.Sprintf(`root = generate_ip(class:%q)`, class)
+func Test_IpV4PrivateA(t *testing.T) {
+	ipType := string(IpV4_PrivateA)
+	mapping := fmt.Sprintf(`root = generate_ip(ip_type:%q)`, ipType)
```
nickzelei, line 192: Can we also have a bloblang test here that provides no inputs to ensure that it always works with defaults?

`mappings := "root = generate_ip()"`

worker/pkg/benthos/transformers/generate_ip_address.go
```diff
@@ -0,0 +1,204 @@
...
+}
+func isReservedIP(ip net.IP) bool {
+	for _, ipRange := range ipv4Ranges {
+		if inRange(ip, ipRange.start, ipRange.end) {
+			return true
+		}
+	}
+	// Check experimental/reserved range (240.0.0.0/4)
+	experimentalStart := net.ParseIP("240.0.0.0")
+	experimentalEnd := net.ParseIP("255.255.255.255")
```
nickzelei, line 176: Could move these to be module level to save having to parse the IP string every time.

nickzelei (approved):
overall, lgtm. nice!

PR #2909, opened 2024-11-05
NEOS-1601: update regex to allow multi word last names

evisdrenova:


worker/pkg/benthos/transformers/transformer_initializer_test.go
```diff
@@ -710,7 +710,7 @@ func Test_InitializeTransformerByConfigType(t *testing.T) {
 		require.NoError(t, err)
 		require.IsType(t, "", result)
 		require.NotEmpty(t, result)
-		require.Regexp(t, `^[A-Z][a-z]+$`, result)
+		require.Regexp(t, `^[A-Z][a-z]+(?:\s[A-Z][a-z]+)*$`, result)
```
nickzelei, line 713: I think you'll also need to update the `GenerateFirstName` tests as well as the `GenerateFullName` tests as those now support spaces.
A full name could possibly now be up to four words I'm guessing.

nickzelei (commented):
I think a few more tests are going to need their regex updated (if they have them?)

worker/pkg/benthos/transformers/transformer_initializer_test.go
evisdrenova: `GenerateFirstName` and `GenerateFullName` actually didn't even have those regexes but I just added them and made them consistent

nickzelei (approved):

PR #2911, opened 2024-11-05
Fix S3 to SQL database types

alishakawaguchi:
Fixes: 
- bit columns
- json columns with string as value

Postgres Broken
- money
- timetz
- all arrays

### DIFF
--- a/internal/benthos/benthos-builder/builders/sql.go
+++ b/internal/benthos/benthos-builder/builders/sql.go
@@ -374,14 +374,21 @@ func (b *sqlSyncBuilder) BuildDestinationConfig(ctx context.Context, params *bb_
 			}
 		}
 
-		columnTypes := []string{}
+		columnTypes := []string{} // use map going forward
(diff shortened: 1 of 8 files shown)


internal/benthos/benthos-builder/builders/sql.go
```diff
@@ -374,16 +374,23 @@ func (b *sqlSyncBuilder) BuildDestinationConfig(ctx context.Context, params *bb_
 			}
 		}
 
-		columnTypes := []string{}
+		columnTypes := []string{} // use map going forward
```
nickzelei, line 377: what does this comment mean?

nickzelei (approved):

internal/benthos/benthos-builder/builders/sql.go
alishakawaguchi: I don't want to continue using column types and column default properties as lists. I want them to be maps where key is column name because I have to do all this extra work to keep them all in the same order as the columns list

PR #2926, opened 2024-11-11
Fix CLI sync postgres uuid being incorrect

alishakawaguchi:


internal/benthos/benthos-builder/builders/sql.go
```diff
@@ -280,13 +280,20 @@ func (b *sqlSyncBuilder) BuildDestinationConfig(ctx context.Context, params *bb_
 		sqlSchemaColMap := getSqlSchemaColumnMap(ctx, params.DestConnection, b.sqlSourceSchemaColumnInfoMap, b.sqlmanagerclient, params.Logger)
 		b.sqlDestinationSchemaColumnInfoMap = sqlSchemaColMap
 	}
+	if len(b.sqlDestinationSchemaColumnInfoMap) == 0 {
```
nickzelei, line 283: It's not clear to me from this context, but does this assume that if the user has marked "init schema", that it has already been initialized. So we are expecting the DB to be hydrated with the schema at this point?

nickzelei (approved):

internal/benthos/benthos-builder/builders/sql.go
alishakawaguchi: I renamed the variable to be more clear. Really that map holds the merged source and destination schema map. So if destination is empty then it relies on the source. The only time it expects the DB to be hydrated with the schema in the destination is for CLI sync where S3 is the source.

nickzelei (approved):

PR #2927, opened 2024-11-11
NEOS-1609: update ip addres function to handle ipv4 and ipv6 byte ranges correctly

evisdrenova:


worker/pkg/benthos/transformers/generate_ip_address.go
```diff
@@ -185,7 +185,32 @@ func generateIPInRange(randomizer rng.Rand, start, end net.IP) string {
 	return int64ToIP(random).String()
 }
 
+// normalizes IPs into the same format
+func normalizeIP(ip net.IP) net.IP {
```
nickzelei, line 189: interesting.

nickzelei (approved):

PR #2935, opened 2024-11-13
NEOS-1616: [MIGRATION] Adds RBAC engine

nickzelei:
This is kind of a large PR but it does a few things here:

- Adds Casbin RBAC engine
- On bootup, upserts existing accounts with their permissions. This is to ensure that all accounts have the latest permission set going forward and for legacy accounts created prior to RBAC being enforced
- On account creation, the following now happens: account policies created, creating user added as account admin
- Adds new RPC method that allows for updating a user's role in the account
- Refactors each RPC service with a new `userdata` service that now encapsulates all of the previous checks we were doing in the `user-account.go` files, as well as ensures the user is _in_ the requesting resource's account
- RBAC is an EE feature and is conditionally enabled if neosync api comes online with a EE or neosync cloud license
- API Keys have now been restricted to account_admins only until we enable fine-grained access to users that is based on their role

Each account now has the following roles:
1. account_admin
2. job_developer
3. job_viewer


alishakawaguchi (approved):

backend/internal/ee/rbac/enforcer/rbac_model.conf
```diff
@@ -11,4 +11,4 @@ g = _, _, _
 e = some(where (p.eft == allow))
 
 [matchers]
-m = g(r.sub, p.sub, r.dom) && r.dom == p.dom && keyMatch2(r.obj, p.obj) && keyMatch2(r.act, p.act)
+m = g(r.sub, p.sub, r.dom) && r.dom == p.dom && keyMatch(r.obj, p.obj) && keyMatch(r.act, p.act)
```
nickzelei, line 14: after more inspection of the functions available, keyMatch is what we actually want.
https://casbin.org/docs/function/

PR #2951, opened 2024-11-18
Fixes REDIS values never being unset, allows disabling backend migrations init container in helm charts

nickzelei:


worker/charts/worker/templates/env-vars.yaml
```diff
@@ -73,17 +73,9 @@ stringData:
     REDIS_MASTER: {{ .Values.redis.master }}
     {{- end }}
 
-    {{- if and .Values.redis .Values.redis.tls .Values.redis.tls.enabled }}
     REDIS_TLS_ENABLED: {{ .Values.redis.tls.enabled | default "false" | quote }}
```
nickzelei, line 76: Seeing strange behavior here where when this isn't provided, the value is never set to false if it was previously set to true.

worker/charts/worker/templates/env-vars.yaml
```diff
@@ -73,17 +73,9 @@ stringData:
     REDIS_MASTER: {{ .Values.redis.master }}
     {{- end }}
 
-    {{- if and .Values.redis .Values.redis.tls .Values.redis.tls.enabled }}
```
nickzelei, line 76: the base values always specified redis.tls so it is safe to remove this.

alishakawaguchi (approved):

PR #2965, opened 2024-11-20
Fixes Postgres money and xml arrays in sync

alishakawaguchi:
cleans up conversion from pgx to go types

### DIFF
--- a/internal/postgres/utils.go
+++ b/internal/postgres/utils.go
@@ -16,16 +16,6 @@ type PgxArray[T any] struct {
 	colDataType string
 }
 
-// func (a *PgxArray[T]) SimpleArray() []any {
-// 	newArray := make([]any, len(a.Elements))
-
-// 	for _, element := range a.Elements {
-
-// 	}
-// 	return newArray
-
-// }
-
 // Handles scanning postgres arrays
 func (a *PgxArray[T]) Scan(src any) error {
 	m := pgtype.NewMap()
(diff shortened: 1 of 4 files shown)


internal/postgres/utils.go
```diff
@@ -20,7 +20,71 @@ type PgxArray[T any] struct {
...
+			ElementType: &pgtype.Type{
+				Name:  "money",
+				OID:   790,
+				Codec: pgtype.TextCodec{},
+			},
+		},
+	})
+
+	// Register UUID types
+	m.RegisterType(&pgtype.Type{
```
nickzelei, line 42: Why do you have to register the uuid type when pgx has a `pgtype.UUID`?

internal/postgres/utils.go
```diff
@@ -20,7 +20,71 @@ type PgxArray[T any] struct {
...
+			ElementType: &pgtype.Type{
+				Name:  "uuid",
+				OID:   2950,
+				Codec: pgtype.TextCodec{},
+			},
+		},
+	})
+
+	// Register XML type
+	m.RegisterType(&pgtype.Type{
```
nickzelei, line 61: Same Q here, they have a `pgtype.XMLCodec`, what is the diff?

nickzelei: Definitely doesnt look like there is an explicit xml scanner but still confused on the UUID one

internal/postgres/utils.go
alishakawaguchi: I'm overriding the codec so that is returns the uuid as a string instead of []byte

internal/postgres/utils.go
alishakawaguchi: same thing I want the xml as a string instead of the xml struct

internal/postgres/utils.go
nickzelei: makes sense

nickzelei: kinda surprised youre able to register it since pgx already registers the uuid type, unless it just overrides it (which I'm guessing is what happens)

alishakawaguchi: yep its just a map so the value for that key gets updated when you register

nickzelei (approved):

PR #2978, opened 2024-11-21
NEOS-1638: fixes table styling

evisdrenova:
This PR updates the table layout and styling to make it aligned with the headers. 

Main things I did here are: 
- fix the column header width and the row cell width to be the same so they scale the same 
- account for the checkbox width
- update the truncated text to actually work. We had a bug where taliwind doesn't recognize partial classnames with dynamic variables in the className prop. You have to use a style prop. Read more [here](https://tailwindcss.com/docs/content-configuration#dynamic-class-names). 
- updated the width of the transformer button to fit in the column header with - this meant removing the "Select" frm the "Select Transformer" button text and just truncated it to "Transformer"

I kept the cells left aligned but we can easily center them if we'd like
Demo
https://www.loom.com/share/70aaf74a230046d8beece3e1c5340ac0?sid=b3ab42b1-3e4a-4357-a107-cb77e809fcaa

### DIFF
--- a/frontend/apps/web/components/TruncatedText.tsx
+++ b/frontend/apps/web/components/TruncatedText.tsx
@@ -31,7 +31,7 @@ export default function TruncatedText(props: Props): ReactElement {
     <TooltipProvider>
       <Tooltip delayDuration={delayDuration}>
         <TooltipTrigger asChild>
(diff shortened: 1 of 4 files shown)


nickzelei (approved):
Nice!! Looks great.

PR #2993, opened 2024-11-26
Neosync Interval and Interval Array type

alishakawaguchi:
This changes interval type in javascript transformer to an object instead of a string


worker/pkg/benthos/neosync_connection_data/neosync_connection_data_input.go
```diff
@@ -189,12 +201,18 @@ func (g *neosyncInput) Read(ctx context.Context) (*service.Message, service.AckF
...
 	}
+
+	registry := neosynctypes.NewTypeRegistry(g.logger)
 	valuesMap := map[string]any{}
 	for col, val := range row {
 		if len(val) == 0 {
 			valuesMap[col] = nil
 		} else {
-			valuesMap[col] = val
+			newVal, err := registry.Unmarshal(val)
```
nickzelei, line 211: so clean

PR #2995, opened 2024-11-26
NEOS-1641: Fixes stale DB Connections during Data Sync Workflow

nickzelei:
This is a pretty large refactor that does a few things:

- Consolidates DB Connection Caching into a single connection manager
- Updates SQL Manager to use ConnectionManager under the hood (part of the connection consolidation)
- Updates the Data Sync Workflow activities to properly release their connection sessions once they are finished with them
- Moves the connection reaper out of the Sync activity and into the startup script
- Centralizes the DB pools for the data sync activity to use a single pool that is shared across all (in-process) activities
- Updates the backend to use the same connection manager logic but to operate in the auto-close mode once a session is released since everything here is transactional
- Updates default connection timeout to be 10s and centralizes
- Updates a bunch of places that hardcoded postgres and pgx to use the module constants


backend/services/mgmt/v1alpha1/connection-data-service/connection-data_test.go
```diff
@@ -1,956 +1,956 @@
...
-		},
-	})
-
-	require.Nil(t, err)
-	require.Len(t, resp.Msg.TableConstraints, 1)
-	require.EqualValues(t, map[string]*mgmtv1alpha1.UniqueConstraint{
-		"public.users": {Columns: []string{"id"}},
-	}, resp.Msg.TableConstraints)
-}
+// import (
```
alishakawaguchi, line 3: did you mean to comment this out?

internal/connection-manager/manager.go
```diff
@@ -0,0 +1,306 @@
...
+		groupConnMap:       map[string]map[string]T{},
+		config:             cfg,
+		shutdown:           make(chan any),
+		mu:                 &sync.Mutex{},
+	}
+}
+
+func (c *ConnectionManager[T]) GetConnection(
+	session SessionInterface,
+	connection ConnectionInput,
```
alishakawaguchi, line 91: might be helpful to have a few debug logs here about when it finds an existing vs when it creates a new one

worker/pkg/workflows/datasync/workflow/workflow_integration_test.go
```diff
@@ -1573,31 +1573,32 @@ func (f *fakeEELicense) IsValid() bool {
 }
 
 func executeWorkflow(
-	t *testing.T,
+	t testing.TB,
```
alishakawaguchi, line 1576: 👍

alishakawaguchi (approved):
Very clean! Left a few comments nothing blocking

internal/connection-manager/manager.go
nickzelei: Good callout. I updated the interface to require the logger and filled it out throughout with more messages.
This forced me to pass it down further anyways and replaced a few instances where I was using the global slog.

internal/testutil/testcontainers/redis/redis.go
```diff
@@ -32,6 +34,10 @@ func (r *RedisTestContainer) Setup(ctx context.Context) (*RedisTestContainer, er
 		"docker.io/redis:7",
 		redis.WithSnapshotting(10, 1),
 		redis.WithLogLevel(redis.LogLevelVerbose),
+		testcontainers.WithWaitStrategy(
+			wait.ForLog("* Ready to accept connections"),
+			wait.ForExposedPort(),
```
nickzelei, line 39: This seems to have fixed the redis container flakiness.

PR #3000, opened 2024-12-02
Wraps temporal logger in slog.Logger for better system plugin

nickzelei:
Also updates a bunch of spots to use the testutil.GetLogger for tests


alishakawaguchi (approved):
🥇

PR #3006, opened 2024-12-03
Nick/neos 1644 wire up job hooks to UI

nickzelei:


frontend/apps/web/app/(mgmt)/[account]/jobs/[id]/layout.tsx
```diff
@@ -195,40 +195,14 @@ interface SidebarNav {
...
-  isMetricsServiceEnabled?: boolean
+  isMetricsServiceEnabled?: boolean,
+  isJobHooksEnabled?: boolean
 ): SidebarNav[] {
   if (!job) {
     return [{ title: 'Overview', href: `` }];
   }
   const basePath = `/${accountName}/jobs/${job.id}`;
 
-  if (isDataGenJob(job) || isAiDataGenJob(job)) {
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/jobs/[id]/layout.tsx:205]: this wasnt actually used and was overridden by the outside nav below this if clause

alishakawaguchi (approved):

evisdrenova (changes requested):
Nice! I think functionality looks good but a few things to clean up on the UI. 

Video notes here: https://www.loom.com/share/254f3220ed714ce1bca040d845fc03fe?sid=982f3f16-cac6-4a75-9883-c493417bc343

Written notes:
1. Standardizing on the dialog height and structure. Currently the new hook and edit hook dialogs are different heights. My vote is use the edit hook height and structure. 
2. Nest the `Enabled` and `Hook Type` selections under the header instead of to the right. 
3. Fix the overflowed scrollbar on the edit dialog to be inside the container. 

PS. i played around with having monaco on the dialog and was able to get the resizer working ( couldn't help myself) using some ref and useEffects - it's not perfect, there is a slight delay when you resize that causes the editor to overflow as it's resizing which could prob be fixed by debouncing the api call to the editor but it's not terrible.

https://www.loom.com/share/dc2aedff50c74ceeb6e44f6cad851fa9?sid=b600b46e-1c45-4d3b-9a7f-a8603806df1c

code:
```typescript
  const editorContainerRef = useRef<HTMLDivElement>(null);
  const [editorWidth, setEditorWidth] = useState<number>(0);

  // last comment in the issue: https://stackoverflow.com/questions/61691616/how-to-make-react-monaco-edtor-responsive
  useEffect(() => {
    if (!editorContainerRef.current) return;

    const resizeObserver = new ResizeObserver((entries) => {
      const entry = entries[0];
      if (entry) {
        setEditorWidth(entry.contentRect.width - 16);
      }
    });

    setEditorWidth(editorContainerRef.current.clientWidth - 16);
    resizeObserver.observe(editorContainerRef.current);

    return () => {
      resizeObserver.disconnect();
    };
  }, []);

    <div
          className="border border-gray-700 p-2 rounded-lg"
          ref={editorContainerRef}
        >
          <Editor
            height="60px"
            width={editorWidth > 0 ? `${editorWidth}px` : '100%'}
            language="sql"
            // value={constructWhere(item?.where ?? '')}
            theme={'vs-dark'}
            // onChange={(e) => onWhereChange(e?.replace('WHERE ', '') ?? '')}
            options={editorOptions}
          />
        </div>
```

evisdrenova (approved):

PR #3009, opened 2024-12-04
Remove default postgres driver and use PGX everywhere

alishakawaguchi:


backend/pkg/sqlmanager/postgres/postgres-manager.go
```diff
@@ -15,6 +15,8 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const PostgresDriver = "postgres"
```
nickzelei, line 18: does this need to be exposed outside of this module?

backend/pkg/dbconnect-config/postgres.go
```diff
@@ -8,9 +8,10 @@ import (
 
 	mgmtv1alpha1 "github.com/nucleuscloud/neosync/backend/gen/go/protos/mgmt/v1alpha1"
 	"github.com/nucleuscloud/neosync/backend/pkg/clienttls"
-	sqlmanager_shared "github.com/nucleuscloud/neosync/backend/pkg/sqlmanager/shared"
 )
 
+const PostgresDriver = "postgres"
```
nickzelei, line 13: does this need to be exposed outside of this module?

backend/pkg/sqlmanager/postgres/postgres-manager.go
nickzelei: In the querybuilder we have a `GoquPostgresDriver`, should be we be using that on 878?

alishakawaguchi: Import cycle 😢 . moved it to the shared sqlmanager. So now we just have PostgresDriver and GoquPostgresDriver

backend/pkg/sqlmanager/shared/types.go
```diff
@@ -99,10 +99,9 @@ func ToConstraintType(constraintType string) (ConstraintType, error) {
 }
 
 const (
-	PostgresDriver        = "pgx"
-	DefaultPostgresDriver = "postgres"
-	MysqlDriver           = "mysql"
-	MssqlDriver           = "sqlserver"
+	PostgresDriver = "pgx"
```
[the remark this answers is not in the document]
alishakawaguchi: I removed all the other postgresdrivers

nickzelei (approved):

PR #3044, opened 2024-12-12
NEOS-1672: Fixes worker error shutdown, k8s helm secret revisions

nickzelei:
- Fixes the worker not properly shutting down on error. This was being caused by not properly calling shutdown on the mongo connection manager
- Adds more logging to the worker startup
- Updates helm secrets to be revisioned based on their shasum. This fixes the issue of secrets not properly being removed from the secret (due to how k8s manages secret merges) which was causing presence issues during startup.
- Removes the `rollMe` from the helm annotations. Now that secrets are re-generated on change, if there are no changes the deployment will not roll for no reason


worker/internal/cmds/worker/serve/serve.go
```diff
@@ -281,14 +282,14 @@ func serve(ctx context.Context) error {
 
 	mongoconnmanager := connectionmanager.NewConnectionManager(mongoprovider.NewProvider())
 	go mongoconnmanager.Reaper(logger)
-	defer sqlconnmanager.Shutdown(logger)
+	defer mongoconnmanager.Shutdown(logger)
```
nickzelei, line 285: oops

alishakawaguchi (approved):

PR #3079, opened 2024-12-23
NEOS-1692:add in transform uuid transformer

evisdrenova:
This feature adds in a Transform UUID transformer. By default, it generates new UUID v4s. You can pass in a seed value to deterministically output UUIDs based on the input UUID. This is useful if you consider UUIDs to be sensitive data but want users to maintain their UUID consistency across tables/databases. So given a seed value, an input UUID will generate the same output UUID. 

Demo:
https://www.loom.com/share/cc48d55850e246a488f90976b90e2b3f?sid=965d7b94-31fd-412b-93de-d90db181bde5


worker/pkg/benthos/transformers/transform_uuid.go
```diff
@@ -0,0 +1,107 @@
...
+	seedBytes := make([]byte, 16)
+
+	// Use the first 8 bytes from the input UUID
+	copy(seedBytes, inputUuid[:8])
+
+	randomInt := randomizer.Float64()
+	binary.LittleEndian.PutUint64(seedBytes[8:], uint64(randomInt))
+
+	// Create a new UUID using SHA1 namespace
+	output := uuid.NewSHA1(uuid.Nil, seedBytes).String()
```
nickzelei, line 104: Do we actually want to put all of the generated uuids in the same Nil namespace? 
The original design we came up with puts the new uuids inside of the original uuid's namespace.

worker/pkg/benthos/transformers/transform_uuid.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	// Generate a deterministic byte slice based on the input UUID and the randomizer
+	seedBytes := make([]byte, 16)
+
+	// Use the first 8 bytes from the input UUID
+	copy(seedBytes, inputUuid[:8])
+
+	randomInt := randomizer.Float64()
+	binary.LittleEndian.PutUint64(seedBytes[8:], uint64(randomInt))
```
nickzelei, line 101: Curious as to why we are going through this whole song and dance to generate the input data when it accepts any arbitrary amount.

In other words, why not just `uuid.NewSha1(inputUuid, []byte(randomizer.Float64()))`

worker/pkg/benthos/transformers/transform_uuid_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	randomizer := rand.New(rand.NewSource(1))
+
+	//checks that the output uuid is the same everytime for given the same input since we're assigning it a specific seed value
+	for i := 0; i < 5; i++ {
+		res := transformUuid(randomizer, randomUuid)
+		checkVars = append(checkVars, *res)
+	}
+
+	fmt.Println("test", checkVars)
```
nickzelei, line 39: I don't think we want this here. If you want to have a test log something, use the test logger `t.Log` - then it will only show up when running go tests in verbose mode.

nickzelei (changes requested):
Left a few comments regarding the algorithm for transforming the uuids

worker/pkg/benthos/transformers/transform_uuid.go
evisdrenova: updated

worker/pkg/benthos/transformers/transform_uuid_test.go
evisdrenova: ah good catch - that was hold over from testing

worker/pkg/benthos/transformers/transform_uuid.go
evisdrenova: we could do that too

worker/pkg/benthos/transformers/transform_uuid.go
```diff
@@ -91,17 +91,13 @@ func transformUuid(randomizer rng.Rand, value string) *string {
...
-	// Use the first 8 bytes from the input UUID
-	copy(seedBytes, inputUuid[:8])
-
 	randomInt := randomizer.Float64()
-	binary.LittleEndian.PutUint64(seedBytes[8:], uint64(randomInt))
 
-	// Create a new UUID using SHA1 namespace
-	output := uuid.NewSHA1(uuid.Nil, seedBytes).String()
+	bytes := make([]byte, 16)
+	binary.LittleEndian.PutUint64(bytes, uint64(randomInt))
```
nickzelei, line 97: If you go into the randomizer you can expose the `Int63` function, then the `Float64` can just go away entirely.

Then it can become `uuid.NewSHA1(inputUuid, []byte(randomizer.Int63()))`

nickzelei (approved):

PR #3084, opened 2025-01-03
Bumps Neosync to v0.5

nickzelei:
This is a larger release that does a number of things:

1. Removes all `@deprecated` fields in the API. These have all been marked as deprecated for some time and are not used by any of the official clients or the App.
2. Removes unused or deprecated methods. The connection data service had a few erroneous methods that have been replaced for some time by better methods.
3. Moves Object Storage connections to be behind an Enterprise license.
4. Fixes many Postgres data types during syncs
5. Introduces the Neosync data type structure that is used for piping data through the system
6. Fixes (and breaks the old way) of syncing S3 data types to allow for better retention of data types during CLI syncs
7. Logs and Metrics are now behind an Enterprise License. This is only for surfacing them, not for exposing them to be read by external systems
8. Updates TypeScript SDK to use the new Connection v2 API. Anyone that is using the TS sdk will have to update their scripts to support the newer method which lacks classes. The docs have been updated appropriately
9. Adds RBAC - API Keys are now admin only (until we add fine grained control)
10. Other minor bug fixes and updates to tests


alishakawaguchi (approved):
🚢

PR #3155, opened 2025-01-16
Adds warnings to jobmapping validator

alishakawaguchi:
Screenshot 2025-01-16 at 2 55 08 PM


backend/pkg/utils/utils.go
```diff
@@ -26,3 +27,27 @@ func DedupeSliceOrdered[T comparable](input []T) []T {
 	}
 	return output
 }
+
+func DedupeSlice[T comparable](input []T) []T {
```
nickzelei, line 31: FWIW, I think we can actualy remove this function entirely and just use [slices.Compact](https://pkg.go.dev/slices#Compact)

backend/pkg/utils/utils.go
```diff
@@ -26,3 +27,27 @@ func DedupeSliceOrdered[T comparable](input []T) []T {
...
+		set[i] = struct{}{}
+	}
+	output := make([]T, 0, len(set))
+	for key := range set {
+		output = append(output, key)
+	}
+	return output
+}
+
+func CompareSlices(slice1, slice2 []string) bool {
```
nickzelei, line 43: I think we can remove this and use [slices.Compare](https://pkg.go.dev/slices#Compare)

backend/protos/mgmt/v1alpha1/job.proto
```diff
@@ -1072,6 +1072,7 @@ message ValidateJobMappingsRequest {
   string connection_id = 3 [(buf.validate.field).string.uuid = true];
   // The list of virtual foreign keys
   repeated VirtualForeignConstraint virtual_foreign_keys = 4;
+  optional JobSource job_source = 5;
```
nickzelei, line 1075: Please add a description of this property.

backend/protos/mgmt/v1alpha1/job.proto
```diff
@@ -1085,6 +1086,13 @@ message ColumnError {
   repeated string errors = 4;
 }
 
+message ColumnWarning {
+  string schema = 1;
+  string table = 2;
+  string column = 3;
+  repeated string warnings = 5;
+}
```
nickzelei, line 1094: Please add property descriptions (see ColumnError)

backend/protos/mgmt/v1alpha1/job.proto
```diff
@@ -1094,6 +1102,7 @@ message ValidateJobMappingsResponse {
   repeated ColumnError column_errors = 1;
   // The database error
   DatabaseError database_errors = 2;
+  repeated ColumnWarning column_warnings = 3;
```
nickzelei, line 1105: please add prop description.

frontend/apps/web/app/(mgmt)/[account]/jobs/util.ts
```diff
@@ -1765,6 +1766,7 @@ export async function validateJobMapping(
       });
     }),
     connectionId: connectionId,
+    jobSource: jobSource,
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/jobs/util.ts:1769]: how is the job source different from the `connectionId`? aren't they the same thing? Could `connectionid` go away in favor of just the job source?

frontend/apps/web/components/jobs/SchemaTable/FormErrorsCard.tsx
```diff
@@ -109,6 +128,9 @@ export default function FormErrorsCard(props: Props): ReactElement {
 function formErrorsToMessages(errors: FormError[]): string[] {
```
nickzelei, line 128: I'm very performance sensitive about things changing on the schema table page. 
There are potentially a lot of form errors that can occur on this page...

It might be more efficient here to have a `formErrorsToMessages` function return an object that splits them.
```
interface FormErrorMesageResponse {
  errors: string[];
  warnings: string[];
}
```
Then loop through them and check the level.
That way you reduce the algorithm from O(2N) to just O(N)

frontend/apps/web/app/(mgmt)/[account]/jobs/[id]/source/components/DataSyncConnectionCard.tsx
```diff
@@ -418,14 +432,24 @@ export default function DataSyncConnectionCard({ jobId }: Props): ReactElement {
...
+    const connection = connections.find((c) => c.id === values.sourceId);
+    const job = data?.job;
     try {
       setIsValidatingMappings(true);
+      let jobsource: JobSource | undefined;
+      if (job && connection) {
+        jobsource = create(JobSourceSchema, {
+          options: toJobSourceOptions(values, job, connection, values.sourceId),
+        });
+      }
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/jobs/[id]/source/components/DataSyncConnectionCard.tsx:445]: seems liek this logic is pretty similar to what is above in the validateMappings function, but it's not using the same conditional logic. ... should it be? they look pretty darn similar.

nickzelei (commented):
nice..left a few comments

frontend/apps/web/app/(mgmt)/[account]/jobs/util.ts
Comment (nickzelei) [reply frontend/apps/web/app/(mgmt)/[account]/jobs/util.ts:1769]: I still have this question

internal/job/jobmapping-validator.go
```diff
@@ -208,7 +207,7 @@ func (j *JobMappingsValidator) ValidateCircularDependencies(
 	}
 
 	for table, deps := range validForeignKeyDependencies {
-		validForeignKeyDependencies[table] = utils.DedupeSlice(deps)
+		validForeignKeyDependencies[table] = slices.Compact(deps)
```
nickzelei, line 210: heck yeah

nickzelei (approved):
lgtm, just the last question around needing the connection id if we also have the job source (maybe that thing doesnt have the conn id?)

frontend/apps/web/app/(mgmt)/[account]/jobs/util.ts
Comment (alishakawaguchi) [reply frontend/apps/web/app/(mgmt)/[account]/jobs/util.ts:1769]: I wanted to keep the validate job mapping endpoint flexible. It doesn't require the jobsource. If you just want to validate the mappings then all you need is mappings and connection_id. Like you could call it with the mappings and destination id and it will tell you all the missing tables

Comment (nickzelei) [reply frontend/apps/web/app/(mgmt)/[account]/jobs/util.ts:1769]: sounds good

PR #3156, opened 2025-01-16
Updates mysql and postgres docs to include notes about database differences

evisdrenova:


docs/docs/connections/mysql.md
```diff
@@ -10,6 +10,11 @@ Neosync is an open-source, developer-first product that allows you to create ano
 
 Neosync supports most Mysql-compatible databases natively using the Mysql connection.
 
+## Things to watch out for
+
+1. Currently Neosync only supports syncing between two physical databases and does not support syncing between two logical databases. This is on the roadmap and will be supported soon.
+2. When syncing across two databases, the databases must have the same name.
```
nickzelei, line 16: This makes sense but am not sure it is BDS or really gets the point across. Someone is going to read this and not truly understand the meaning.
It would most likely benefit if you gave a more explicit example where you showed two URLs and explain that even though this is specified, in Neosync we still surface all of the databases.

Let's lay it out.

Physical Mysql Server (server1):
  - Database1 (virtual database, or seen as "schema" in Neosync)
      - table1 (table) 
  - Database2 (virtual database, or seen as "schema" in Neosync)
      - table2 (table)  

Even though the user can specify this in the URL, Neosync currently ignores this an still surfaces all of the known databases that it finds. This was to optimize for the ability to sync all of your potential databases from one physical machine down to another physical machine (in the same way that we do for Postgres)

docs/docs/connections/postgres.md
```diff
@@ -13,6 +13,11 @@ Postgres is one of the most commonly used databases in the world and Neosync nat
 
 The following guide will show you how to configure and test your Postgres Connection.
 
+## Things to watch out for
+
+1. Neosync can sync two logical Postgres databases (for ex. within the same Docker container) but the schema names do have to be the same.
+2. When syncing across two databases, the databases must have the same name.
```
nickzelei, line 19: This is not correct or really get the point across.

Postgres has virtual databases and then within those has schemas.
Neosync can connect to the same _physical_ postgres database but have them be filtered by their virtaul database across two neosync connections. We can sync between those, but the _schema_ and _tables_ will have to be the same.

let's lay it out.

Physical Postgres Server (server1):
  - Database1 (virtual database)
      - public (schema)
          - table1 (table)  
  - Database2 (virtual database)
      - public (schema)
          - table1 (table)

Someone can create two Neosync connections where one is connected to `Database1` and the other `Database2` - and Neosync can sync between them.  

This is different from Mysql because when you specify the database in the postgres url, that connection is incapable of querying the other database, it may only query across schemas within it.

nickzelei (changes requested):

docs/docs/connections/postgres.md
nickzelei: One additional note, for what it's worth, you definitely _can_ connect to a database in postgres and still query across from them. But we don't do that. The query effectively looks like this: `select * from database1.public.table1` but if it's omitted, it just defaults to the database that you connected to. 

But mysql fell ill to this because it isn't triple nested (and is extremely uncommon to support this in Postgres). If we had built mysql first, we most likely would have built it in such a way that mysql was scoped properly.

nickzelei: we built our job mappings to have a `schema, table, column` with no mapping. So Mysql doesn't really work well in this scenario because it is not also triple nested. we'll have to think thru how to support that in our job mappings.

nickzelei (approved):

PR #3163, opened 2025-01-17
NEOS-1712: Adds support for JS in Anon Api

nickzelei:
* Adds support for javascript in anonymization api
* Refactors JS VM out of Benthos processor for easier re-use


internal/javascript/functions/functions.go
```diff
@@ -0,0 +1,140 @@
...
+	MetaGet(name string) (any, bool)
+	MetaSetMut(name string, value any)
+}
+
+type Ctor func(r Runner) Function
+
+type FunctionDefinition struct {
+	namespace string
+	name      string
+	// params    []*FunctionParam
```
alishakawaguchi, line 31: what's this

alishakawaguchi (approved):

internal/javascript/functions/functions.go
nickzelei: forgot to remove

PR #3169, opened 2025-01-20
Setup continuous benchmarking

alishakawaguchi:
Updates generators to create transformer benchmarks
Adds github action for continuous benchmarking
  - compares main branch benchmarks to PR branch benchmarks using benchstat
  - leaves comment on PR with geomean results


nickzelei (approved):
This is awesome, let's chat more about the PR comment output and how we can make it more meaningful.

I'm guessing the current printout is the total overall benchmark?
Would be great if we could get a tabular rendered view of each function being benchmarked.

PR #3185, opened 2025-01-23
NEOS-1727, NEOS-1676, NEOS-1728: Adds docs for hooks, rbac, and more

nickzelei:


docs/src/css/overrides.css
```diff
@@ -319,3 +319,7 @@ header h2 a {
 .breadcrumbs__link {
   @apply py-1 px-2;
 }
+
+.markdown > blockquote {
```
nickzelei, line 323: adds the same margin styling that <p> tags have.

evisdrenova (approved):

PR #3202, opened 2025-01-29
Fixes PG Data Type Arrays on schema init, re-enables integration tests in CI

nickzelei:
Also fixes integration tests not fully running in the new format (doh)

The int test fix has exposed critical race errors that were being uncaught due to integration tests not running properly

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -131,7 +131,7 @@ jobs:
 
       - name: Run Integration Tests
         run: |
-          go test -race -timeout 1800s -coverprofile=integration-coverage.out -covermode=atomic -run TestIntegrationTestSuite ./...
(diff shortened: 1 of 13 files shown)


.github/workflows/go.yml
```diff
@@ -131,7 +131,7 @@ jobs:
 
       - name: Run Integration Tests
         run: |
-          go test -race -timeout 1800s -coverprofile=integration-coverage.out -covermode=atomic -run TestIntegrationTestSuite ./...
+          go test -race -timeout 1800s -coverprofile=integration-coverage.out -covermode=atomic ./...
```
nickzelei, line 134: 🤦

evisdrenova (approved):

PR #3208, opened 2025-01-31
fixing circular dependency run order

nickzelei:


backend/pkg/table-dependency/circular-dependencies.go
```diff
@@ -101,14 +101,15 @@ func FindCircularDependencies(dependencies map[string][]string) [][]string {
 	var result [][]string
```
nickzelei, line 101: It's unclear to me if this update is actually better or not.
I was working with Claude on this and it was saying that having the visited here was causing some paths to not be fully visited, however I may need to think about this some more to see if this is actually just less efficient code.

Either way, all of the exists tests pass and I've added more that also pass.
The code is now also simpler too so it's unclear. Could be worth also benchmarking too.

backend/pkg/table-dependency/table-dependency.go
```diff
@@ -336,7 +336,8 @@ func processCycles(
 			if isTableInCycles(cycles, fkTable) {
 				if len(fkCols.NullableColumns) > 0 {
 					updateConfig.appendDependsOn(fkTable, fkCols.NullableColumns)
-				} else {
+				}
+				if len(fkCols.NonNullableColumns) > 0 {
```
nickzelei, line 340: This is the main bugfix here that I think is causing issues with circular dependencies.

This was resulting in a table that has a non-nullable dependency from being added as a dependency in the run order.
It is being added as a necessary foreignKey in the runconfig, but not as a dependency.
This results in both tables being synced at the same time, which causes FK constraint errors.

internal/benthos/benthos-builder/builders/processors_test.go
```diff
@@ -23,7 +23,7 @@ func Test_buildProcessorConfigsJavascript(t *testing.T) {
 		Config: &mgmtv1alpha1.TransformerConfig{
 			Config: &mgmtv1alpha1.TransformerConfig_TransformJavascriptConfig{
 				TransformJavascriptConfig: &mgmtv1alpha1.TransformJavascript{
-					Code: `return "hello " + value;`,
+					Code: `return "hello " + value + " " + input.extra;`,
```
nickzelei, line 26: this is a drive by change after working with a discord user. I noticed we don't actually have any tests that ensure the input works in the JS transformer processor.

worker/pkg/workflows/datasync/activities/sync/activity_test.go
```diff
@@ -191,7 +191,7 @@ pipeline:
   threads: 1
   processors:
     - mutation: |
-        root.name = fake("first_name")
+        root.name = generate_first_name()
```
nickzelei, line 194: This is also a drive by change. I was attempting to replace bento wtih benthos core and it doesn't look like `fake` is available anymore. We don't expose this to end-users anyways so I am updated this to be one of our functions.

alishakawaguchi (approved):

PR #3221, opened 2025-02-05
Fix assignment to entry in nil map when providing attributes in mysql connection string

### LINKED ISSUE #3220: Assignment to entry in nil map when adding attributes to sql connection string
## Bug description
When i want to add a new sql connection, and set attributes, i get a panic (see screenshot)

Image

## How to reproduce

1. Add a new SQL connection, with the following format `mysql://<user>:<pwd>@<host>:<port>/<schema>?<attributes>`
2. Try `Test connection`
3. You will get an error back.

## Environment
Local docker containers

## Additional context
The error happens in https://github.com/nucleuscloud/neosync/blob/main/backend/pkg/dbconnect-config/mysql.go#L96. The code wants to assign values to the not initialised `cfg.Params` map.

Yarn-e:
I've added a test to prove the issue, and fixed it.

Fixes #3220

### DIFF
--- a/backend/pkg/dbconnect-config/mysql.go
+++ b/backend/pkg/dbconnect-config/mysql.go
@@ -93,9 +93,13 @@ func NewFromMysqlConnection(
 			}
 			cfg.MultiStatements = true
 			cfg.ParseTime = parseTime
-			for k, values := range uriConfig.Query() {
-				for _, value := range values {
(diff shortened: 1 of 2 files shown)


nickzelei (approved):
Thank you for fixing this and for providing a test!

PR #3243, opened 2025-02-10
NEOS-1739: adds webhook workflow

nickzelei:
- adds the webhook workflow to the worker

Follow up PRs
- Wires up event publishers
- Wires up UI


compose/temporal/.env
```diff
@@ -1,9 +1,10 @@
 COMPOSE_PROJECT_NAME=temporal
 CASSANDRA_VERSION=3.11.9
-ELASTICSEARCH_VERSION=7.16.2
+ELASTICSEARCH_VERSION=7.17.27
```
nickzelei, line 3: 7.16 was magically breaking after I updated docker and macos. This version seems to have fixed it.

PR #3252, opened 2025-02-12
Adds environment variable for table sync max currency limit

alishakawaguchi:


worker/charts/worker/templates/_helpers.tpl
```diff
@@ -126,4 +126,7 @@ NEOSYNC_CLOUD: {{ .Values.neosyncCloud.enabled | default "false" | quote }}
 {{- if and .Values.ee .Values.ee.license }}
 EE_LICENSE: {{ .Values.ee.license | quote }}
 {{- end }}
+{{- if .Values.tableSync.maxConcurrency }}
+TABLESYNC_MAX_CONCURRENCY: {{ .Values.tableSync.maxConcurrency }}
+{{- end }}
```
nickzelei, line 131: The if wrapper is not required because you've set it as a default value in the values.yaml

worker/pkg/workflows/datasync/workflow/workflow.go
```diff
@@ -623,3 +623,11 @@ func getAccountStatusTimerDuration() time.Duration {
 	}
 	return time.Duration(envtime) * time.Second
 }
+
+func getTableSyncMaxConcurrency() int {
+	maxConcurrency := viper.GetInt("TABLESYNC_MAX_CONCURRENCY")
+	if maxConcurrency == 0 {
```
nickzelei, line 629: Best to change this to `<=0` to prevent people from setting it to something like `-1`

nickzelei (approved):

PR #3266, opened 2025-02-15
Adds Resumable Sync for higher stability of syncing large datasets

alishakawaguchi:


backend/pkg/dbschemas/sql/postgresql/queries/system.sql
```diff
@@ -680,3 +679,29 @@ GROUP BY
...
+  JOIN pg_catalog.pg_class AS idx ON idx_meta.indexrelid = idx.oid
+  -- Unnest the index key attribute numbers along with their ordinal positions
+  JOIN unnest(idx_meta.indkey) WITH ORDINALITY AS key_info(attnum, ordinality) ON true
+  -- Join to get the column attributes corresponding to the index keys
+  JOIN pg_catalog.pg_attribute AS col ON col.attrelid = tbl.oid AND col.attnum = key_info.attnum
+WHERE ns.nspname = ANY(sqlc.arg('schema')::TEXT[])
+  AND idx_meta.indisunique = true
+  -- Exclude indexes automatically created by constraints
+  AND NOT EXISTS (
+       SELECT 1 FROM pg_catalog.pg_constraint AS cons WHERE cons.conindid = idx.oid
```
nickzelei, line 705: Hm, the thing that might get us here is we saw with $customer that they had a unique index that we were excluding from our index query because it was created by a constraint. However, that constraint no longer existed, but the index was around..and there were FK's that were utilizing this unique index for their references.

I'm wondering now if it would actually be valuable to include these indices here. 
That way, if the original constraint exists, great the algo will use that. If not, then it falls back on the dangling index. @alishakawaguchi

alishakawaguchi: yeah that makes sense 👍

alishakawaguchi: fixed

nickzelei (approved):

PR #3274, opened 2025-02-17
NEOS-1747, NEOS-1754: Updates to on conflict and other async sql mechs

nickzelei:


frontend/apps/web/yup-validations/transformer-validations.ts
```diff
@@ -147,7 +147,7 @@ const generateInt64Config = Yup.object().shape({
 
 const generateStringPhoneNumberConfig = Yup.object().shape({
   min: getBigIntValidator({
-    range: [8, 12],
+    range: [8, 14],
```
nickzelei, line 150: NEOS-1747

worker/pkg/benthos/sql/input_sql_raw.go
```diff
@@ -41,6 +39,7 @@ func RegisterPooledSqlRawInput(
 			if err != nil {
 				return nil, err
 			}
+			// return service.AutoRetryNacksToggled(conf, input)
```
nickzelei, line 42: I added this here because this is what benthos has configured today. Although I'm leaving it commented out because we have never used this and it is supposed to retry downstream errors endlessly, which I don't think we really want our system to do today. Just leaving this here as a reminder if we ever want it.

worker/pkg/benthos/sql/input_sql_raw.go
```diff
@@ -150,13 +146,9 @@ func (s *pooledInput) Connect(ctx context.Context) error {
 	s.dbMut.Lock()
 	defer s.dbMut.Unlock()
 
-	if s.db != nil {
```
nickzelei, line 153: I removed this because Connect will retry until a non-error is returned. But if connect fails to retrieve the records, we will have set the DB, so the row input will be empty and Read will return with no input, which may be a source of bugs for us. What needs to happen is this needs to retry until Connect returns successfully.

worker/pkg/benthos/sql/input_sql_raw.go
```diff
@@ -150,13 +146,9 @@ func (s *pooledInput) Connect(ctx context.Context) error {
...
-		return nil
-	}
-
 	db, err := s.provider.GetDb(ctx, s.connectionId)
 	if err != nil {
-		return nil
+		return err
```
nickzelei, line 151: whoops

alishakawaguchi (approved):

PR #3285, opened 2025-02-20
Runconfigs V2 - adds stability, removes complexity

alishakawaguchi:
Simplifies run config and query builder logic.

For circular dependencies and subsets removes complicated logic around optimal path. Now it determines update configs based on tables nullable foreign keys. A separate update config is created for each nullable foreign key on the table.

Moves subset joining logic into run config builder because we create update configs for subset tables now. It computes the shortest path from every child table to all parent subsets. This reduces unnecessary and complicated joins


internal/benthos/benthos-builder/builders/sql.go
```diff
@@ -135,29 +134,21 @@ func (b *sqlSyncBuilder) BuildSourceConfigs(ctx context.Context, params *bb_inte
...
-	// used in destination config to determine if foreign key violations should be skipped
-	for table, querymap := range tableRunTypeQueryMap {
-		b.isNotForeignKeySafeSubsetMap[table] = make(map[tabledependency.RunType]bool)
-		for runtype, q := range querymap {
-			b.isNotForeignKeySafeSubsetMap[table][runtype] = q.IsNotForeignKeySafeSubset
-		}
-	}
-
-	configs, err := buildBenthosSqlSourceConfigResponses(logger, ctx, b.transformerclient, groupedTableMapping, runConfigs, sourceConnection.Id, tableRunTypeQueryMap, groupedColumnInfo, filteredForeignKeysMap, colTransformerMap, job.Id, params.JobRunId, b.redisConfig, primaryKeyToForeignKeysMap)
+	configs, err := buildBenthosSqlSourceConfigResponses(logger, ctx, b.transformerclient, groupedTableMapping, runConfigs, sourceConnection.Id, configQueryMap, groupedColumnInfo, filteredForeignKeysMap, colTransformerMap, job.Id, params.JobRunId, b.redisConfig, primaryKeyToForeignKeysMap)
```
nickzelei, line 151: guess the logic for not FK safe subset was moved elsewhere.

worker/pkg/select-query-builder/subset.go
```diff
@@ -0,0 +1,43 @@
...
+func BuildSelectQueryMap(
+	driver string,
+	runConfigs []*runcfg.RunConfig,
+	subsetByForeignKeyConstraints bool,
+	pageLimit int,
+) (map[string]*sqlmanager_shared.SelectQuery, error) {
+	qb := NewSelectQueryBuilder("public", driver, subsetByForeignKeyConstraints, pageLimit)
+	querymap := map[string]*sqlmanager_shared.SelectQuery{}
+	for _, cfg := range runConfigs {
+		query, _, pageQuery, isNotForeignKeySafe, err := qb.BuildQuery(cfg)
```
nickzelei, line 29: confused as to where the `isNotForeignKeySafe` is defined.

alishakawaguchi: I should probably rename that. It really just tells me when a query is a subset query and needs to turn skip fk violations when inserting or updating. It gets determined when building the select query.

nickzelei (approved):

PR #3292, opened 2025-02-21
wip on thinking thru workflow dsl

nickzelei:


worker/pkg/workflows/datasync/workflow2/workflow.go
```diff
@@ -0,0 +1,68 @@
...
+	eelicense license.EEInterface
+}
+
+func New(eelicense license.EEInterface) *Workflow {
+	return &Workflow{
+		eelicense: eelicense,
+	}
+}
+
+func (w *Workflow) Workflow(ctx workflow.Context, req *WorkflowRequest) (*WorkflowResponse, error) {
```
nickzelei, line 27: This is a stub refactor, but the better approach will ultimately be the DSL configuration.

PR #3314, opened 2025-03-03
fixes postgres deadlock detection

nickzelei:


backend/pkg/sqlretry/dbtx_retry.go
```diff
@@ -107,6 +107,35 @@ func (r *RetryDBTX) BeginTx(ctx context.Context, opts *sql.TxOptions) (*sql.Tx,
 	return retry(ctx, operation, r.config.getRetryOpts)
 }
 
+func (r *RetryDBTX) RetryTx(ctx context.Context, opts *sql.TxOptions, fn func(*sql.Tx) error) error {
```
nickzelei, line 110: This is more useful than the BeginTx above as it handles the transaction lifecycle for you.

backend/pkg/sqlretry/deadlock.go
```diff
@@ -29,8 +30,8 @@ func isMysqlDeadlockError(err error) bool {
 	if err == nil {
 		return false
 	}
-	mysqlErr, ok := err.(*mysql.MySQLError)
-	if ok {
+	var mysqlErr *mysql.MySQLError
+	if errors.As(err, &mysqlErr) {
```
nickzelei, line 34: we know that this already works as we are seeing proper deadlock detection errors for mysql users in our logs, but this is an extra safeguard if these errors ever become wrapped.

backend/pkg/sqlretry/deadlock.go
```diff
@@ -40,10 +42,17 @@ func isPostgresDeadlock(err error) bool {
 	if err == nil {
 		return false
 	}
-	pqErr, ok := err.(*pq.Error)
-	if ok {
+
+	var pgErr *pgconn.PgError
```
nickzelei, line 46: fail

alishakawaguchi (approved):

PR #3351, opened 2025-03-11
Moves init schema to separate workflow + creates schema manager

alishakawaguchi:
Creating the schema manager makes it easier to only update mysql to do incremental schema updates and let sqlserver and postgres stay the same


worker/pkg/workflows/datasync/workflow/workflow.go
```diff
@@ -419,6 +409,44 @@ func execRunJobHooksByTiming(ctx workflow.Context, req *jobhooks_by_timing_activ
...
+			MaximumAttempts: 1,
+		},
+		HeartbeatTimeout: 1 * time.Minute,
+	}
+	for _, destinationId := range destinationIds {
+		logger.Info("scheduling Schema Initialization workflow for execution.", "destinationId", destinationId)
+		siWf := &schemainit_workflow.Workflow{}
+		var wfResult schemainit_workflow.SchemaInitResponse
+		err := workflow.ExecuteChildWorkflow(workflow.WithChildOptions(ctx, workflow.ChildWorkflowOptions{
+			WorkflowID:    workflow_shared.BuildChildWorkflowId(jobRunId, "init-schema", workflow.Now(ctx)),
```
nickzelei, line 430: worth adding the destination id into the workflow id so that it is unique. yea we are using the time here but it might still cause dupes if it happens quick enough.

nickzelei (approved):
left one comment regarding workflow id for child

PR #3382, opened 2025-03-18
NEOS-1774: Adds dropping and updating triggers to Mysql Init Schema

alishakawaguchi:


internal/schema-manager/mysql/mysql.go
```diff
@@ -114,6 +116,20 @@ func getDatabaseDataForSchemaDiff(
 		return nil
 	})
 
+	trigmu := sync.Mutex{}
+	errgrp.Go(func() error {
```
nickzelei, line 120: is this running in a loop or is anything else other than this goroutine writing to the triggers slice?? if not I think the mutex is not needed.

internal/schema-manager/mysql/mysql.go
```diff
@@ -114,6 +116,20 @@ func getDatabaseDataForSchemaDiff(
...
+		tabletriggers, err := db.Db().GetSchemaTableTriggers(ctx, tables)
+		if err != nil {
+			return fmt.Errorf("failed to retrieve database table triggers: %w", err)
+		}
+		for _, tabletrigger := range tabletriggers {
+			trigmu.Lock()
```
nickzelei, line 126: if the mutex is actually needed, it might be better to just lock it before the for loop and then defer unlock right after. let the for loop hydrate the map and then exit, unlocking the mutex.

internal/schema-manager/mysql/mysql.go
```diff
@@ -127,10 +143,17 @@ func getDatabaseDataForSchemaDiff(
 			if err != nil {
 				return fmt.Errorf("failed to retrieve  database table constraints for schema %s: %w", schema, err)
 			}
-			for key, tableconstraint := range tableconstraints {
-				mu.Lock()
-				constraints[key] = tableconstraint
-				mu.Unlock()
+			for _, tableconstraint := range tableconstraints {
+				for _, nonFkConstraint := range tableconstraint.NonForeignKeyConstraints {
+					mu.Lock()
```
nickzelei, line 148: same Q's here. idk if the mutex is needed unless more than one thing in a different goroutine is writing to this map.
If it _is_ needed, it's excessive to lock/unlock within this for loop I think. especially given you are just overwriting the map contents and not checking if there is a collision.

nickzelei (approved):
overall lgtm, let some minor comments regarding concurrent syncs

PR #3412, opened 2025-03-26
NEOS-1787: add auto map support to mysql

nickzelei:


backend/pkg/dbschemas/sql/mysql/queries/system.sql
```diff
@@ -214,6 +215,7 @@ SELECT
    c.TABLE_NAME AS table_name,
    c.COLUMN_NAME AS column_name,
    c.COLUMN_TYPE AS data_type,
```
nickzelei, line 217: this is kinda weird. the above query we are using c.data_type but this one we are using c.column_type which is different. I didnt want to update it at the expense of potentially causing a bux.

backend/pkg/dbschemas/sql/mysql/queries/system.sql
```diff
@@ -214,6 +215,7 @@ SELECT
...
+   c.COLUMN_TYPE AS column_type, -- same as data_type but includes things like the length, or set/enum information
```
nickzelei, line 218: @alishakawaguchi - This column_type properly tracks mysql `set` and `enum` values. So I'm guessing that the mysql schema drift changes won't track if this changes. I created a ticket for this: NEOS-1791

alishakawaguchi (approved):

PR #3424, opened 2025-03-28
adds more logic for conflicts, updates to use our sql adapter

nickzelei:
this should hopefully squash any further permission denied errors.


go.mod
```diff
@@ -48,6 +47,7 @@ require (
 	github.com/microsoft/go-mssqldb v1.8.0
 	github.com/neilotoole/slogt v1.1.0
 	github.com/nucleuscloud/go-antlrv4-parser v0.0.0-20240830015744-041b46c70aa5
+	github.com/nucleuscloud/sql-adapter v0.0.0-20250327235651-6f5a05a7cedc
```
nickzelei, line 50: This ultimately is the fix as I updated the insert logic to just on conflict do nothing instead of conflict.

internal/ee/rbac/policy.go
```diff
@@ -388,20 +408,26 @@ func (r *Rbac) EnforceAccount(
 	return nil
 }
 
-func setPolicy(e casbin.IEnforcer, policy []string) error {
+type setPolicyResult struct {
+	DidConflict bool
+}
+
+func setPolicy(e casbin.IEnforcer, policy []string) (*setPolicyResult, error) {
```
nickzelei, line 415: the on conflict logic is kind of dead for now but leaving it in here as a final safety mechanism

alishakawaguchi (approved):

PR #3474, opened 2025-04-08
NEOS-1808: job run activity timeline status filter not working

nickzelei:
- Makes timeline status more useable. Now can select or deselect all
- Fixes nested button error with multi select tooltip
- Updates activity table columns to use better sorting and allows filters now

image


backend/services/mgmt/v1alpha1/job-service/runs.go
```diff
@@ -273,6 +273,10 @@ func (s *Service) getEventsByWorkflowId(
 			isRunComplete = true
 		case enums.EVENT_TYPE_WORKFLOW_EXECUTION_TIMED_OUT:
 			isRunComplete = true
+		case enums.EVENT_TYPE_WORKFLOW_EXECUTION_TERMINATED:
```
nickzelei, line 276: these were causing the run events to still be refreshed even though the job was complete

alishakawaguchi (approved):

PR #3477, opened 2025-04-08
NEOS-1714 add hash transformer

alishakawaguchi:


worker/pkg/benthos/transformers/transform_hash.go
```diff
@@ -0,0 +1,183 @@
...
+	return hashToString(hasher, valueStr)
+}
+
+func valueToStringForHash(value any) (string, error) {
+	switch v := value.(type) {
+	case string:
+		return v, nil
+	case []byte:
+		return string(v), nil
+	// todo: handle more types
```
nickzelei, line 171: Would it be difficult to fill in more types here? Ideally this thing works for any type and we can allow it for any or most data types. Unclear how this will appear on the FE atm though.

nickzelei (approved):
LGTM, left a comment regarding supported types.

worker/pkg/benthos/transformers/transform_hash.go
nickzelei: I'm moreso wondering fi there is a way to effectively cast any input type to either []byte or string. Whichever is the most convenient for passing in to a hash function. You can hash anything if it's a []byte...

worker/pkg/benthos/transformers/transform_hash.go
```diff
@@ -0,0 +1,183 @@
...
+	default:
+		return "", fmt.Errorf("invalid value type: %T", value)
+	}
+}
+
+func hashToString(hasher hash.Hash, input string) (string, error) {
+	_, err := hasher.Write([]byte(input))
```
nickzelei, line 178: maybe it's better to make the input a `[]byte` then we just convert the `value any` to a []bte, then we support everything.

PR #3478, opened 2025-04-08
Updates run page to show new schema reconciliation errors

alishakawaguchi:


frontend/apps/web/app/(mgmt)/[account]/runs/[id]/components/JobRunActivityErrors.tsx
```diff
@@ -35,6 +35,20 @@ function parseUint8ArrayToInitSchemaReport(
     return null;
   }
 }
+
+function parseUint8ArrayToReconcileSchemaReport(
+  data: Uint8Array
+): InitSchemaReport[] | null {
```
Comment (nickzelei) [inline frontend/apps/web/app/(mgmt)/[account]/runs/[id]/components/JobRunActivityErrors.tsx:41]: just curious but why as an array if you are always returning a static report?

Comment (alishakawaguchi) [reply frontend/apps/web/app/(mgmt)/[account]/runs/[id]/components/JobRunActivityErrors.tsx:41]: to make it compatible with the old schema init report component

